PR #52, opened 2017-08-28
Use AtomicLong for object ID generation.

SvetlinZarev:
The long type provides wide enough namespace for unique IDs
and the AtomicLong type provides faster way to generate unique
object IDs than the UUID class.

### DIFF
--- a/api/api-facades/api-core/src/main/java/org/eclipse/dirigible/api/v3/core/JavaFacade.java
+++ b/api/api-facades/api-core/src/main/java/org/eclipse/dirigible/api/v3/core/JavaFacade.java
@@ -43,9 +43,7 @@ public class JavaFacade {
 					return json;
 				}
 				// non primitive result - add to the context
-				String resultUuid = generateUuid();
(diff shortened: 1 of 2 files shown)


SvetlinZarev:
Some numbers from multi threaded JMH benchmark to show the difference: 
Using 32 concurrent threads: 

    Benchmark                   Mode  Cnt      Score     Error   Units
    UUIDGeneration.atomicLong  thrpt   10  40354,823 ± 500,333  ops/ms
    UUIDGeneration.uuid        thrpt   10   1644,589 ± 109,979  ops/ms

Single threaded: 

    Benchmark                   Mode  Cnt      Score     Error   Units
    UUIDGeneration.atomicLong  thrpt   10  17230,560 ± 137,689  ops/ms
    UUIDGeneration.uuid        thrpt   10   1748,589 ±  32,502  ops/ms

As you can see the UUID based approach:
* is 10 time slower than AtomicLong for the single threaaded case
* does not scale at all with the number of cores

Issue #764, opened 2021-03-17
[SQL] Enhance exists method to accept artefact type


delchev:
Enhance exists method to accept artefact type

PR #778, opened 2021-03-26
Fix/764 Enhance exists method to accept artefact type 

Wildflowerr:
COMMITTERS: please include labels on each PR. Labels are listed here: https://github.com/eclipse/dirigible/labels
-->

### What does this PR do?
Before hdbsequence could not be updated because method exists would always return false due to the fact prev implementation would deduce that every artefact is a table and SELECT COUNT (*) won't work as a legit check for hdbsequence existence.

https://github.com/eclipse/dirigible/issues/764


ThuF:
It seems that the "first author" of the commit can't be recognized/validated for ECA: https://accounts.eclipse.org/legal/eca/validation/63432

The first commit could be amended and signed from the author with the `-s` option, or just delete this PR and create new one with signed commits.

Wildflowerr:
Close due unsigned commit.

PR #844, opened 2021-04-26
Removed redundant slash that was preventing the copying of the war

BorisNen:
Signed-off-by: Nenchovski <26414287+BorisNen@users.noreply.github.com>

### What does this PR do?
Removes an unnecessary slash for accessing the target directory and copying ROOT.war from desktop-all to CATALINA_HOME. The slash prevented the war copying on MacOS.

None


ThuF (approved):
+1

Issue #846, opened 2021-04-26
[IDE Workspace] File with extension .xsodata can not be deleted


antoanetamh:
**Describe the bug**
File can not be deleted from the workspace.

**To Reproduce**
1. Create file with extension .xsodata
2. Right Click and choose Remove
3. File still exist.

**Expected behavior**
File should be deleted

**Solution:**
Add headers: { 'Dirigible-Editor': 'Workspace' } when remove files

Issue #845, opened 2021-04-26
[IDE Workspace] - When creating files .xsodata file cannot be loaded


antoanetamh:
**Describe the bug**
The .xsodata GET request returns data in format
`{"d":{"EntitySets":["Customers","Orders","CustomersWithoutKey","CustomersWithAllKeys","CustomersWithGenID","orderView","orderViewKeyGen"]}}`

The code in renameNode method of worspace.js expect response in format as:
`{"name":"uu.xssodata","path":"/workspace/toni/uu.xssodata","type":"file","contentType":"text/plain"}`
This is causing the incorrect building of the node JS object.

![image](https://user-images.githubusercontent.com/44926767/116118141-add33400-a6c5-11eb-99f7-230daf4cc089.png)

![image](https://user-images.githubusercontent.com/44926767/116118260-c6434e80-a6c5-11eb-9f6b-dd61f3d4ee03.png)

**To Reproduce**
1. Go Workspace
2. Create file with extension .xsodata
3. Double click over the new created file
4. Welcome page is open. The file can not be open or edit

**Expected behavior**
The file should be open for edit.

**Solution**
Call the refresh method after renameNode method.

PR #848, opened 2021-04-28
fix issue with create and delete of .xsodata

antoanetamh:
https://github.com/eclipse/dirigible/issues/846
https://github.com/eclipse/dirigible/issues/845


ThuF (approved):
+1

Issue #850, opened 2021-04-29
Escaping table name for MySQL not working


Wildflowerr:
If you want to create a table with the following XS Classic definition:

```txt
table.schemaName = "DBADMIN";
table.temporary = true;
table.tableType = COLUMNSTORE;
table.loggingType = NOLOGGING;
table.columns = [
 {name = "ID"; sqlType = INTEGER; unique= false; length = 40; nullable = false; comment = "hello"; defaultValue = 20; precision = 2; scale = 15;},
 {name = "NAME"; sqlType = VARCHAR; length = 20; nullable = false; },
 {name = "JOB"; sqlType = VARCHAR; length = 20; nullable = false;},
 {name = "NUMBER"; sqlType = INTEGER; length = 20; nullable = false; defaultValue = 444;}];
table.indexes =  [
 {name = "MYINDEX"; unique = false; indexColumns = ["ID"];},
 {name = "MYINDEX10"; unique = true; indexColumns = ["JOB", "NAME"];}];
table.primaryKey.pkcolumns = ["ID"];
table.description = "test table test";
```
SqlFactory creates query of the following form:
```sql
CREATE TABLE "hdbtable-itest::SamplePostgreXSClassicTable" ( ID INTEGER  DEFAULT 20  NOT NULL , NAME VARCHAR (20) NOT NULL , JOB VARCHAR (20) NOT NULL , NUMBER INTEGER  DEFAULT 444  NOT NULL , 
PRIMARY KEY ( ID ), CONSTRAINT "MYINDEX" UNIQUE ( ID ), 
CONSTRAINT "MYINDEX10" UNIQUE ( JOB , NAME ))
```
Such SQL syntax leads to a db error: 
```txt
You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '"hdbtable-itest::SamplePostgreXSClassicTable"
```
According to mysql documentation, table names should be escaped with `` ` `` instead of `` " `` symbol.

PR #860, opened 2021-05-13
PostgreCreateViewBuilder & MySQLCreateViewBuilder adapted for xsk hdbview

Wildflowerr:
Introduces implementation logic for these [integration tests](https://github.com/SAP/xsk/issues/168).

It appears that for postgresql, only table/view names should be escaped solely. Other properties should not be escaped.
The following sql is a good working example:
```sql
CREATE VIEW "hdbview-itest::SamplePostgreXSClassicViewppp" AS SELECT T1.Column2 FROM "public"."acme.com.test.tables::MY_TABLE1" AS T1 LEFT JOIN "public"."acme.com.test.views::MY_VIEW1" AS T2 ON T1.Column1 = T2.Column1;
```

For MySQL, the proper [escaping symbol](https://github.com/eclipse/dirigible/issues/850) is considered.


ThuF (approved):
+1

PR #887, opened 2021-06-07
Session API - added language and invocation count methods. Improved previous methods.

BorisNen:
### What does this PR do?
Adds UserFacade methods that are necessary for the implementation of the language member and the getInvocationCount method in session API in the xsk project. Improves previous Session API methods. Adds a counter to the getSession method inside HttpSessionFacade.

This PR is related to two different issues inside the xsk project - https://github.com/SAP/xsk/issues/218 and https://github.com/SAP/xsk/issues/219

#### Documentation
Documentation will be provided inside the xsk wiki.

### DIFF
--- a/api/api-facade/api-http/src/main/java/org/eclipse/dirigible/api/v3/http/HttpSessionFacade.java
+++ b/api/api-facade/api-http/src/main/java/org/eclipse/dirigible/api/v3/http/HttpSessionFacade.java
@@ -47,11 +47,7 @@ public class HttpSessionFacade implements IScriptingFacade {
 			if (request != null) {
 				HttpSession httpSession = request.getSession(true);
 				Integer count = (Integer)httpSession.getAttribute(INVOCATION_COUNT);
-				if (count == null) {
-					count = 1;
-				} else {
-					count++;
-				}
+				count = count == null ? 1 : ++count;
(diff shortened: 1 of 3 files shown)


api/api-facade/api-http/src/main/java/org/eclipse/dirigible/api/v3/http/HttpSessionFacade.java
```diff
@@ -44,7 +45,15 @@ static final HttpSession getSession() {
...
 			HttpServletRequest request = (HttpServletRequest) ThreadContextFacade.get(HttpServletRequest.class.getCanonicalName());
 			if (request != null) {
-				return request.getSession(true);
+				HttpSession httpSession = request.getSession(true);
+				Integer count = (Integer)httpSession.getAttribute(INVOCATION_COUNT);
+				if (count == null) {
+					count = 1;
+				} else {
+					count++;
+				}
```
ThuF, line 54: Instead if-else you can try this one:
```java
count = count == null ? 1 : count ++;
```

BorisNen: It is replaced in the next commit.

ThuF (approved):
+1

PR #897, opened 2021-06-15
Added view and layout settings

StanZGenchev:
Signed-off-by: StanZGenchev <stanislav.genchev@sap.com>

### What does this PR do?

Adds support for tab and layout settings from GoldenLayout. This enables per view and per layout configuration.


ThuF (approved):
+1

Issue #913, opened 2021-06-24
[Database] SQL Builder - Issue with Creation of PK


antoanetamh:
When table is created with added data for Pk on column level and primarykey(), the assembled SQL in bulded incorrectly with double PRIMARY KEY("ID","ID2") ,PRIMARY KEY("ID","ID2")

SqlFactory.getNative(new HanaSqlDialect())
				.create()
				.columnTable("CUSTOMERS")
				.columnInteger("ID", **true**, false, false)
				.columnInteger("ID2", **true**, false, false)
				.columnVarchar("FIRST_NAME", 20, false, true, true)
				.columnVarchar("LAST_NAME", 30, false, true, false)
				**.primaryKey(new String[]{"ID", "ID2"}**)

PR #914, opened 2021-06-24
fix issue with dublicate primarykey in SQL assemble string

antoanetamh:
fixed https://github.com/eclipse/dirigible/issues/913


ThuF (approved):
+1

Issue #922, opened 2021-06-30
[IDE] Expose 'set current workspace' functionality as an API call


StanZGenchev:
**Describe the issue**
Currently it's not possible to change the selected workspace from code. The only way is to use the workspace UI drop-down. Exposing this functionality as an API call will help other perspectives such as `xsk-ide-perspective`.

**Expected behavior**
The API call should change the currently selected workspace in local storage but it should not refresh any views.

ThuF:
Fixed with #923

PR #923, opened 2021-06-30
Fixes #920 and #922

### LINKED ISSUE #920: [IDE] Expand 'Open perspective' functionality
**Describe the issue**
Currently the 'Open perspective' functionality in ide-core can open a perspective in a new window.
The only way to open a perspective in the current window, is to use the menu or sidebar.

**Expected behavior**
There should be a 'ide-core.openPerspective' call similar to 'openView' accept it will open a perspective instead of a view and all parameters should be passed as a query string in the URL.

StanZGenchev:
Signed-off-by: StanZGenchev <stanislav.genchev@sap.com>

### What does this PR do?

Exposes 'set current workspace' functionality and adds new 'open perspective' API calls.
Fixes #920 and #922


ThuF (approved):
+1

PR #937, opened 2021-07-06
odata-core java compatibility set to 1.8 to enable more use cases

mit-ko:
### What does this PR do?
increases the JDK compatibility of odata-core-module

Fix a bug with wrong SQL generation on update/create of odata2 entities


mit-ko:
will create a new one. This one has a manual edit, which github cannot recognize

PR #938, opened 2021-07-06
Fixed a bug with ODATA update/create of entities

mit-ko:
Fixed an issue creating wrong sql for post/put/merge of OData2 entities

Added OData2 tests for post/put/merge

Added compatibility for java 1.8 for the OData2 core module

### What does this PR do?

Fixed a bug that prevents post/put/merge of OData2 entities

### DIFF
--- a/modules/odata/odata-core/pom.xml
+++ b/modules/odata/odata-core/pom.xml
@@ -21,8 +21,8 @@
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>${maven.compiler.plugin.version}</version>
 				<configuration>
-					<source>${maven.compiler.source}</source>
-					<target>${maven.compiler.target}</target>
+					<source>1.8</source>
+					<target>1.8</target>
 					<debug>true</debug>
 					<debuglevel>lines,vars,source</debuglevel>
 				</configuration>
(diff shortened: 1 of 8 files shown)


mit-ko:
Would it be also possible to merge it in the latest 5.x branch?
Thanks in advance

modules/odata/odata-core/pom.xml
```diff
@@ -21,12 +21,20 @@
 				<artifactId>maven-compiler-plugin</artifactId>
 				<version>${maven.compiler.plugin.version}</version>
 				<configuration>
-					<source>${maven.compiler.source}</source>
-					<target>${maven.compiler.target}</target>
+					<source>1.8</source>
```
delchev, line 24: Do we need to set it explicitly to 1.8? Currently it is 11 which supports the features used below

modules/odata/odata-core/pom.xml
```diff
@@ -21,12 +21,20 @@
...
+					<target>1.8</target>
```
delchev, line 25: Same here - it is 11 already

delchev (changes requested):

modules/odata/odata-core/pom.xml
mit-ko: The thing is that when a module is compiled with java 11 it cannot be used with projects that run on java 1.8

modules/odata/odata-core/pom.xml
mit-ko: Same reason here

ThuF (approved):
+1

delchev (approved):

Issue #999, opened 2021-07-29
[OData] Bug with POST / PUT / PATCH objects and related Entity Id's


avandenhoek:
**Describe the bug**
Integration Patterns has a relation to two other entities Styles / Domains. If I try to POST / P it does not process the related entities properly. 

```json
{
    "CreatedBy": "SAP",
    "Name": "IP-Postman",
    "Domain": {
        "Id": "b4dc3e21-bacb-4fed-aa02-70273525fb76"
    },
    "Style": {
        "Id": "ed9a8b52-ca8e-11e9-a32f-2a2ae2dbcce4"
    }
}
```

The error is as follows (internal 500):
```json
{
    "error": {
        "code": "Internal Server Error",
        "message": {
            "lang": "en",
            "value": "org.postgresql.util.PSQLException: ERROR: null value in column \"domain_id\" of relation \"integrationpatterns\" violates not-null constraint\n  Detail: Failing row contains (45680e59-e7dd-4cf7-a3ea-eb514a38174a, 2021-07-29 15:20:19.597917, SAP, 2021-07-29 15:20:19.597917, null, IP-Postman, null, null)."
        }
    }
}
```

Somehow it does not translate it to the Internal Foreign key DOMAIN_ID /  STYLE_ID.

If I do a PATCH / PUT, the result is different, the request body is:
```json
{
    "Id": "42af5a91-3357-4e98-9a3f-3308dcfb7beb",
    "Name": "IP-Postman Updated",
    "Style": {
        "Id": "42af5a91-3357-4e98-9a3f-3308dcfb7xyz"
    },
    "Domain": {
        "Id": "42af5a91-3357-4e98-9a3f-3308dcfb7beb"
    }
}
```

It gives a 204 No content, but does not seem to have done anything with the Style and Domain. The Name field is updated correctly.

Metadata:
```xml
<?xml version='1.0' encoding='UTF-8'?>
<edmx:Edmx xmlns:edmx="http://schemas.microsoft.com/ado/2007/06/edmx" Version="1.0">
    <edmx:DataServices m:DataServiceVersion="1.0" xmlns:m="http://schemas.microsoft.com/ado/2007/08/dataservices/metadata">
        <Schema Namespace="com.sap.intas.backend.entity.domain.methodology" xmlns="http://schemas.microsoft.com/ado/2008/09/edm">
            <EntityType Name="Domains">
                <Key>
                    <PropertyRef Name="Id"/>
                </Key>
[...]

PR #1004, opened 2021-07-30
Add support for Create/Update Entity with Reference id

mit-ko:
### What does this PR do?
Improvement for the OData Create operation

Fixes in the OData2 layer

- Accepts inline reference entity on create

- Accepts inline reference entity on update

- Fixes connection leak issues in the OData Core

- Various code improvements on the OData layer

Fixes
https://github.com/eclipse/dirigible/issues/999


ThuF:
@mit-ko, as the https://github.com/eclipse/dirigible/issues/999 issue wasn't automatically resolved, I just wan't to double check whether we are expecting more changes on it, or I can close it?

PR #1154, opened 2021-09-20
Fix put

mit-ko:
### What does this PR do?
Fix a put with invalid keys


mit-ko:
replaced by https://github.com/eclipse/dirigible/pull/1155

PR #1170, opened 2021-09-24
Move beforeEval after require initialisation

tdermendjiev:
beforeEval should be called between require/module initialisation and module loading / script evaluation


vmutafov (approved):

modules/engines/engine-javascript/src/main/java/org/eclipse/dirigible/engine/js/api/IJavascriptEngineExecutor.java
```diff
@@ -62,4 +62,7 @@
 	@Override
 	public Object executeServiceModule(String module, Map<Object, Object> executionContext) throws ScriptingException;
 
+	@Override
+	public Object evalModule(String module, Map<Object, Object> executionContext) throws ScriptingException;
```
vmutafov, line 66: Is there a reason for this method to be overridden here? Isn't it the same as in the parent interface? Also, if we keep it, please add a (non-Javadoc) as on the method above

PR #1206, opened 2021-10-20
feat: configure exec facade to log based on the configuration

vmutafov:
Solution for: https://github.com/SAP/xsk/issues/452


modules/commons/commons-config/src/main/java/org/eclipse/dirigible/commons/config/Configuration.java
```diff
@@ -448,7 +448,8 @@ public static void setSystemProperty(String key, String value) {
 			"DIRIGIBLE_PRODUCT_COMMIT_ID",
 			"DIRIGIBLE_PRODUCT_TYPE",
 			"DIRIGIBLE_INSTANCE_NAME",
-			"DIRIGIBLE_SPARK_CLIENT_URI"
+			"DIRIGIBLE_SPARK_CLIENT_URI",
+			"DIRIGIBLE_EXEC_DISABLE_COMMAND_LOGGING"
```
ThuF, line 452: Let's align with the other environment variables:

`DIRIGIBLE_EXEC_DISABLE_COMMAND_LOGGING` -> `DIRIGIBLE_EXEC_COMMAND_LOGGING_ENABLED`

modules/engines/engine-command/src/main/java/org/eclipse/dirigible/engine/command/processor/CommandEngineExecutor.java
```diff
@@ -221,6 +224,11 @@ public String executeCommandLine(String commandLine, Map<String, String> forAddi
 		return result;
 	}
 
+	private boolean shouldLogCommand() {
+		String shouldEnableLogging = Configuration.get("DIRIGIBLE_EXEC_DISABLE_COMMAND_LOGGING");
+		return Boolean.parseBoolean(shouldEnableLogging);
```
ThuF, line 229: You could add static import for the DIRIGIBLE_EXEC_COMMAND_LOGGING_ENABLED constant and update the block as follows:
```java
return Boolean.parseBoolean(Configuration.get(DIRIGIBLE_EXEC_COMMAND_LOGGING_ENABLED, "false"));
```

vmutafov: I prefer to keep it on two lines as it looks a bit more readable to me and there is no need for the default "false" as parseBoolean handles nulls as well.

As for if it should log or not log, I think it's better to not log by default

ThuF (approved):
+1

Issue #1271, opened 2021-11-10
.hdbtable tableType is not taken into account


dpanayotov:
Issue cloned from https://github.com/SAP/xsk/issues/756

Using the following table definition doesn't take into account table.tableType property value "ROWSTORE" and create COLUMN table instead of ROW table.

```
table.schemaName = "DBADMIN";
table.temporary = true;
**table.tableType = ROWSTORE;**
table.loggingType = NOLOGGING;
table.columns = [
	{name = "Col1"; sqlType = NVARCHAR; nullable = false; length = 20; comment = "dummy comment";},
	{name = "Col2"; sqlType = INTEGER; nullable = false;},
	{name = "Col5"; sqlType = NVARCHAR; nullable = false; length = 20; defaultValue = "Defaultvalue";},
    {name = "Col66"; sqlType = NVARCHAR; nullable = true; length = 20; defaultValue = "Defaultvalue";},
    {name = "Col77"; sqlType = NVARCHAR; nullable = true; length = 20; defaultValue = "Defaultvalue";}];

table.primaryKey.pkcolumns = ["Col1", "Col2","Col5"];
```

LiliyaLazarova:
Assign to me

Issue #1270, opened 2021-11-10
.hdbtable indeces type in not taken into account


dpanayotov:
Issue cloned from https://github.com/SAP/xsk/issues/410

If we have the following hdbtable definition:
```text
table.schemaName = "DBADMIN";
table.temporary = true;
table.tableType = COLUMNSTORE;
table.loggingType = NOLOGGING;
table.columns = [
	{name = "Col1"; sqlType = NVARCHAR; nullable = false; length = 20; comment = "dummy comment";},
	{name = "Col2"; sqlType = INTEGER; nullable = false;},
	{name = "Col5"; sqlType = NVARCHAR; nullable = false; length = 20; defaultValue = "Defaultvalue";},
    {name = "Col66"; sqlType = NVARCHAR; nullable = true; length = 20; defaultValue = "Defaultvalue";},
    {name = "Col77"; sqlType = NVARCHAR; nullable = true; length = 20; defaultValue = "Defaultvalue";}];
table.indexes =  [
	{name = "BB112"; unique = true; order = DSC; indexColumns = ["Col2"];},
	{name = "BB212"; unique = true; order = DSC; indexType = B_TREE; indexColumns = ["Col1"];}];
table.primaryKey.pkcolumns = ["Col1", "Col2","Col5"];
```

the SQL to be generated is:
```sql
CREATE COLUMN TABLE "DBADMIN"."p1::file" ( "Col1" NVARCHAR (20) NOT NULL , "Col2" INTEGER  NOT
 NULL , "Col5" NVARCHAR (20) DEFAULT 'Defaultvalue'  NOT NULL , "Col66" NVARCHAR (20) DEFAULT 'Defaultvalue'  , "Col77" NVARCHAR (20) DEFAULT 'Defaultvalue'  , PRIMARY KEY("Col1" , "Col2" , "Col5") , CONSTRAINT "BB112" UNIQUE ( "Col2" ), CONSTRAINT "BB212" UNIQUE ( "Col1" ))
```
As you can deduce by the snippet, the index type is not considered at all. All indeces have been created as normal constraints. This might be a **limitation** on our side.

LiliyaLazarova:
Assign to me

PR #1289, opened 2021-11-18
adding support for Hana Row table and indexes creation

LiliyaLazarova:
Signed-off-by: LiliyaLazarova <liliya.nikolaeva.lazarova@gmail.com>

### What does this PR do?

- Creation of ROW table type is now possible for Hana DB
- Creation of unique and non-unique indexes is now possible. Index properties are now taken into account as well

.hdbtable indeces type in not taken into account #1270
.hdbtable tableType is not taken into account #1271


ThuF (approved):
+1

Issue #1314, opened 2021-11-30
[Releng] Istio option for helm chart 


krasimirdermendzhiev:
Add Istio option for helm chart.

krasimirdermendzhiev:
Done with https://github.com/eclipse/dirigible/commit/0589c6bf5852f31644e6ac13a5e47ec093918b4b

PR #1315, opened 2021-11-30
add istio option for helm chart

krasimirdermendzhiev:
### What does this PR do?
Add istio option for helm chart.

#1314


ThuF (approved):
+1

PR #1322, opened 2021-12-03
refactor: replace var with let/const

iwolkov:
This PR:
Replace `var` with `let`/`const` in api-database.
Fix test for nextval.js


ThuF (approved):
+1

PR #1343, opened 2021-12-10
[DO NOT MERGE] Synonym target objects search now works only with a given schema

### LINKED ISSUE #1296: [XSOdata] Searching for DB artefact should be only in default user schema
Issue cloned from https://github.com/SAP/xsk/issues/824

When OData service is defined the logic behind is checking whether the related DB artefacts exist, this check should be done only in the default schema and in the PUBLIC schema in case the artefact is a synonym.

related classes: XSKODataParser in XSK and DBMetadataUtil in Dirigible

BorisNen:
Resolves #1296 

Changelog:
getSynonymTargetObjectMetadata inside DBMetadataUtil will search by synonym name and schema name


ThuF:
@BorisNen please update the PR, once it's ready to be merged.

PR #1376, opened 2021-12-21
Set APPLICATIONUSER when getting DB connection

### LINKED ISSUE #1375: [Analytic Privileges] Set APPLICATIONUSER in db session
Issue cloned from https://github.com/SAP/xsk/issues/898

### Background
In APs for row based filtering so far the SESSION_USER was used which is the logged in user. In HANA Cloud this resolves to the technical user with which the connection is established. This means that checks should be adjusted to use a different session variable. HANA Cloud defines `APPLICATIONUSER` in the [SAP Help Portal](https://help.sap.com/viewer/c1d3f60099654ecfb3fe36ac93c121bb/2021_3_QRC/en-US/a16678c085bf475f832c10b2629f3997.html)

### Target
When obtaining a connection use `setClientInfo` with key `APPLICATIONUSER` and value to be the name of the logged in user.
If no user is logged in set the value to `ANONYMOUS`.
This needs to work with both `$.db.getConnection()` and `$.hdb.getConnection()`.
See the Java api for [setClientInfo](https://docs.oracle.com/javase/7/docs/api/java/sql/Connection.html#setClientInfo(java.lang.String,%20java.lang.String))

d-velev:
closes #1375 , closes SAP/xsk#898

### DIFF
--- a/api/api-facade/api-database/pom.xml
+++ b/api/api-facade/api-database/pom.xml
@@ -52,7 +52,13 @@
 			<artifactId>dirigible-database-dynamic</artifactId>
 			<version>7.0.0-SNAPSHOT</version>
 		</dependency>
-	</dependencies>
+        <dependency>
+            <groupId>org.eclipse.dirigible</groupId>
+            <artifactId>dirigible-api-facade-security</artifactId>
+            <version>7.0.0-SNAPSHOT</version>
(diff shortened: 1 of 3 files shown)


ThuF (changes requested):
Refactoring of accessing database product name as discussed

PR #1382, opened 2021-12-23
expand loader indicator area

ivanvolkoff:
Expand loader indicator area on the toolbar.


StanZGenchev:
Canceled workflows as the change is minimal and non-breaking.

Issue #1363, opened 2021-12-17
[API] Expose Git api


tdermendjiev:
Create javascript api with common git operations.

ThuF:
@tdermendjiev I've created a separate issue for the Git API Documentation:
- https://github.com/eclipse/dirigible/issues/1415

PR #1390, opened 2022-01-04
Add more git methods to Git Facade and js api

tdermendjiev:
### What does this PR do?
Implements most of the git operations in the Git Facade.

#1363


api/api-facade/api-tests/src/main/resources/META-INF/dirigible/utils/v4/git/local.js
```diff
@@ -0,0 +1,67 @@
+/*
```
ThuF, line 1: It seems that there’s an issue with the test:

```java
Tests in error: 
65176
  runSuite(org.eclipse.dirigible.engine.js.graalvm.service.GraalVMApiSuiteTest): TypeError: invokeMember (getGitRepositories) on org.eclipse.dirigible.api.v4.git.GitFacade failed due to: Arity error - expected: 2 actual: 1
65177

65178
Tests run: 6, Failures: 0, Errors: 1, Skipped: 0
```

ThuF (changes requested):
@tdermendjiev there’s an issue with the test

delchev (changes requested):
Can we use namespace `git` or `scm/git` for this API instead of `utils`?

api/api-facade/api-git/pom.xml
```diff
@@ -44,9 +44,15 @@
...
         <dependency>
             <groupId>org.eclipse.dirigible</groupId>
-            <artifactId>dirigible-core</artifactId>
+            <artifactId>dirigible-core-git</artifactId>
             <version>7.0.0-SNAPSHOT</version>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.dirigible</groupId>
+            <artifactId>dirigible-commons-test</artifactId>
+            <version>6.1.11</version>
```
ThuF, line 53: Is there a specific reason to reference version **6.1.11** instead of **7.0.0-SNAPSHOT**. Pointing to a specific version could lead in unwanted behaviour after some time and few version increases.

ThuF (changes requested):
Question about a dependency version

ThuF (approved):

PR #1403, opened 2022-01-07
feat: add initial code for watchable typescript compilation

vmutafov:
This PR adds initial support for a watchable TypeScript compiler. 
The provided API uses the `tsc` compiler executable and invokes it with the `-w` flag. This flag starts the compiler in watch mode and this gives us the opportunity to incrementally compile users' projects instead of running the `tsc` compiler on each file save. This improves compilation speed a lot.
Although this approach would be great for development purposes, maybe we still need to hook in the publishing synchronizer in order to guarantee the project is cleanly compiled. Some other things that need to be done are keeping better track of spawned `tsc` compilers, their compilation results, improving the generated `index.d.ts` to reuse the Dirigible `.d.ts` files, and also testing if the generated `tsconfig.json` is the most optimal one.


vmutafov:
@delchev  @ThuF could we have a meeting to brainstorm these things together next week as I have some questions about the publishing synchronizers in the Dirigible?

PR #1404, opened 2022-01-07
User schema search for DB artefacts

### LINKED ISSUE #1296: [XSOdata] Searching for DB artefact should be only in default user schema

BorisNen:
Resolves #1296 

Changelog:
1. Dirigible no longer supports synonym search logic, which is HANA specific
2. Created a default metadata provider and an API when searching for DB artefacts


modules/engines/engine-odata/src/main/java/org/eclipse/dirigible/engine/odata2/transformers/DBMetadataUtil.java
```diff
@@ -249,27 +249,4 @@ public String getArtifactSchema(String artifactName, String[] types) throws SQLE
...
-
-                    targetObjectMetadata.put(ISqlKeywords.KEYWORD_TABLE, name);
-                    targetObjectMetadata.put(ISqlKeywords.KEYWORD_SCHEMA, schema);
-                    targetObjectMetadata.put(ISqlKeywords.KEYWORD_TABLE_TYPE, type);
-                }
-            }
-        }
-
-        return targetObjectMetadata;
-    }
```
ThuF, line 274: @MarinHadzhiev, @delchev, @BorisNen Should we merge it as proposed in the PR, or should we wait for this to be implemented as part of the HANA dialect?

ThuF (changes requested):
This comment (https://github.com/eclipse/dirigible/pull/1404#pullrequestreview-846824102) should be clarified before merging the PR.

modules/engines/engine-odata/src/main/java/org/eclipse/dirigible/engine/odata2/transformers/DBMetadataUtil.java
MarinHadzhiev: @ThuF We have a meeting Monday morning to discuss the implementation option with SQL dialect. Based on the effort estimation will decide whether to merge this one or rework it. Let's keep it open for now.

Issue #651, opened 2020-11-25
[Releng] Fix the logs location in the Docker deployments


ThuF:
Set the logs location to `<Catalina-Home-Location>/logs`, as of now the Dirigible logs _(e.g. _**dirigible-xxx-yyy.log**)_ files are stored at the  `<Catalina-Home-Location>` level.

The log files location configuration should be set through `DIRIGIBLE_OPERATIONS_LOGS_ROOT_FOLDER_DEFAULT` environment variable, where the default value is `../logs`.

This change should be applicable only for the **Docker** releases as:
- Cloud Foundry (All and Runtime)
- Kyma (All and Runtime)
- Server (Local Docker)

ThuF:
@vmutafov this is the relevant issue in Eclipse Dirigible, regarding this PR in XSK: https://github.com/SAP/xsk/pull/958

PR #1419, opened 2022-01-10
fix: reconfigure logback.xml file appenders path

vmutafov:
Fixes: https://github.com/eclipse/dirigible/issues/651

This PR configures the logback.xml files to use the following environment variables for logs output:

1. DIRIGIBLE_LOGS_DIR
2. CATALINA_BASE
3. CATALINA_HOME

The variable to be used, if several of those exist, would be the first one found in the same order as in the list above.
If none of those variables exist, logs would be output the same way as before this PR - relative to the folder where the tomcat was started.


ThuF (approved):
+1

PR #1420, opened 2022-01-10
Revert setClientInfo in WrappedDataSource - to be fixed

d-velev:
This reverts the changes done in https://github.com/eclipse/dirigible/pull/1376 as and exception is thrown in some cases, it will be fixed in another PR

### DIFF
--- a/api/api-facade/api-database/src/main/java/org/eclipse/dirigible/api/v3/db/DatabaseFacade.java
+++ b/api/api-facade/api-database/src/main/java/org/eclipse/dirigible/api/v3/db/DatabaseFacade.java
@@ -641,10 +641,6 @@ public class DatabaseFacade implements IScriptingFacade {
 		}
 		Connection connection = dataSource.getConnection();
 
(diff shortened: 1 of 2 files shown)


BorisNen (approved):

api/api-facade/api-database/src/main/java/org/eclipse/dirigible/api/v3/db/DatabaseFacade.java
```diff
@@ -641,10 +641,6 @@ public static final Connection getConnection(String databaseType, String datasou
 		}
 		Connection connection = dataSource.getConnection();
 
-		if(((WrappedDataSource) dataSource).getDatabaseName().equals(DATABASE_NAME_HDB)) {
-            connection.setClientInfo("APPLICATIONUSER", UserFacade.getName());
-		}
```
ThuF, line 646: Do we need to revert the whole thing or only these 3 lines?

ThuF (changes requested):
Check comments

Issue #1439, opened 2022-01-12
[Core] Parent stack objects not visible


vmutafov:
When creating new stacks in the ThreadContextFacade we should make parent objects visible in the newly created stack. This is needed as otherwise new stacks don't see core objects like Tomcat's current Request/Response objects and others. 

This problem could be easily observed for example if a Flowable process is started from a REST endpoint and a certain task in the process tries to create a new Workspace with the WorkspaceFacade. It internally creates a Workspace at the wrong registry path as it doesn't see the current logged user and assumes it's an anonymous user.

PR #1440, opened 2022-01-12
feat: keep parent stack objects in new stacks

vmutafov:
Fixes: https://github.com/eclipse/dirigible/issues/1439

This PR sets newly created stacks to have their parent's stack as their own one in order for the `get` methods to have access to their parent's objects. 

Although this could be implemented by modifying the `get` method to lookup into parent stacks and use a bit less memory, it would be computationally heavier as the `get` methods are called often.


ThuF (approved):
+1

PR #1442, opened 2022-01-13
Initial support for OData batching

### LINKED ISSUE #843: [OData] Batch support
Add support for `$batch` operations.
batch READ operations are supported.

### LINKED ISSUE #1282: [OData] Support for transaction handling in SQL Processor
**Is your feature request related to a problem? Please describe.**
As per [OData specification](https://www.odata.org/documentation/odata-version-2-0/batch-processing/) it is possible to define change sets within batch requests. A change set can contain multiple requests and shall be interpreted as an atomic unit of work (i.e. a transaction).

Currently the Dirigible Backend does a connection management per request on its own, e.g. https://github.com/eclipse/dirigible/blob/v6.1.0/modules/odata/odata-core/src/main/java/org/eclipse/dirigible/engine/odata2/sql/processor/AbstractSQLProcessor.java#L449. Additionally it does not trigger any transaction commands (e.g. commit, rollback), thus relying that the datasource is configured with autocommit. For changes in a changeset that would mean, that each change is executed isolated in an own transaction, i.e. not atomic.

**Describe the solution you'd like**
We would like to have some kind of transaction management inside the backend. Either custom code should be called to retrieve connections and do commits/rollback or the solution should be somehow like the Transactional annotation within Java EE/Spring: If there is no connection/context available, do the transaction management on your own, if there is context available reuse the connection.

mit-ko:
### What does this PR do?
Provides support for transactions and batching #1282 #843 

Closes #1282 
Closes #843

Note: If you want to use transactions for customer operations, you can use a custom data source that uses only one connection, set auto-commit false on the connection perform the requests and then commit the transaction. The class org.eclipse.dirigible.engine.odata2.sql.utils.SingleConnectionDataSource  can be used for that purpose in custom code.
This approach is very similar to how the changesets in the batch support are implemented. 

Support for batching and transactions. Added unit tests.

### DIFF
--- a/modules/odata/odata-core-test/src/main/java/org/eclipse/dirigible/engine/odata2/sql/AbstractSQLProcessorTest.java
+++ b/modules/odata/odata-core-test/src/main/java/org/eclipse/dirigible/engine/odata2/sql/AbstractSQLProcessorTest.java
@@ -11,7 +11,9 @@
  */
 package org.eclipse.dirigible.engine.odata2.sql;
 
+import static org.apache.olingo.odata2.api.commons.ODataHttpMethod.GET;
 import static org.easymock.EasyMock.expect;
+import static org.junit.Assert.assertEquals;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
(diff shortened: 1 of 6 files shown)


ThuF (approved):
+1

PR #1443, opened 2022-01-14
fix: handle relative imported scripts from parent folder having dirig…

vmutafov:
Handle cases when a JS file imports another JS file from a sibling folder and the latter imports a `@dirigible` package.


ThuF (approved):
+1

vmutafov:
Fixes: https://github.com/eclipse/dirigible/issues/1446

PR #1444, opened 2022-01-14
fix: add modules.json and add test for sibling imports with dirigible imports

vmutafov:
Fixes: https://github.com/eclipse/dirigible/issues/1447

This PR adds:
- Test for the case of import of a file from a sibling folder that has a `@dirigible` import inside of it
- Add `modules.json` for the `api-bpm` module
- Add empty `bpm.d.ts` file for the `api-bpm` module (this and other `d.ts` files would be updated at a later point)


ThuF (approved):
+1

PR #1448, opened 2022-01-14
Fixes wrong results in the OData in few scenarios where skip/top are used with order by

### LINKED ISSUE #1243: [OData] Sorting with pagination and expand not working
**Describe the bug**
When  using orderBy parameter together with pagination (skip and top) as well as expand, the sorting is not working. It is sorted only block-wise, while the initial sorting is not working.

Guess the issue is related to here https://github.com/eclipse/dirigible/blob/master/modules/odata/odata-core/src/main/java/org/eclipse/dirigible/engine/odata2/sql/builder/SQLQueryBuilder.java#L136-L138

**To Reproduce**
It is important to do the order by already on the preselection of IDs. Deleting the "if" in this case works, but then leads to other issues when doing an orderBy on a association as the join is missing in the first query. 

E.g. query: GET /SomeEntity?$skip=0&$top=10&$orderby=SomeRelation/Name%20asc&$expand=SomeRelation

**Expected behavior**
The sorting is working.

**Additional context**
Dirigible version 6.1.0

mit-ko:
### What does this PR do?
Fixes inconsistency in the OData sql generation in few scenarios where skip/top are used with order by

Fixes an issue with order by Closes #1243


ThuF (approved):
+1

PR #1449, opened 2022-01-17
Fix getGitRepositories returning projects without git repos

tdermendjiev:
GitFacade.getGitRepositories should return only projects with initialized repositories.


ThuF (approved):
+1

PR #1462, opened 2022-01-24
Mail configuration provider

BorisNen:
This PR is related to https://github.com/SAP/xsk/issues/899

Changelog:
1. MailFacade's default instance of the MailClient class now fetches all properties through the mail provider service
2. The mail provider service can be selected through an env variable - DIRIGIBLE_MAIL_CONFIG_PROVIDER
3. Added missing mail env variables to Configuration


modules/mail/mail-env-config/src/main/java/org/eclipse/dirigible/mail/env/EnvMailConfigProvider.java
```diff
@@ -0,0 +1,93 @@
...
+        addValue(properties, key, envKey, null);
+    }
+
+    private static void addValue(Properties properties, String key, String envKey, String defaultValue) {
+        String value = Configuration.get(envKey);
+        if (value != null) {
+            properties.put(key, value);
+        } else if (defaultValue != null) {
+            properties.put(key, defaultValue);
+        }
```
ThuF, line 91: Could we use here `Configuration.setIfNull()`?

ThuF (commented):
@BorisNen +1, Looks good to me, just one small comment, before merging the PR.

modules/mail/mail-env-config/src/main/java/org/eclipse/dirigible/mail/env/EnvMailConfigProvider.java
BorisNen: Configuration.setIfNull() is a void method that sets key-value pairs inside Configuration itself.
Here, we want to check if we should put the returned or default value inside the Properties instance.

ThuF: Agreed, thanks for the update! 😄

PR #1482, opened 2022-01-26
feat: support importing zips into a given path

vmutafov:
Fixes: https://github.com/eclipse/dirigible/issues/1483

This PR adds support for importing zips into a given path which would be treated as relative to the Repository root path.
The new endpoint is a POST to `/project?path=/some/path/relative/to/the/repository/root`


ThuF (approved):
+1

PR #1487, opened 2022-01-27
feat: add new process executor and change ExecFacade implementation

vmutafov:
Related to: https://github.com/eclipse/dirigible/issues/1491 and https://github.com/SAP/xsk/issues/1120

Currently, the ExecFacade uses a process executor that combines the error and out streams. This is not desired in some cases. The current process executor API is not easily extendable and would need some refactoring to support this. 

This PR proposes a new API for executing command-line programs with the idea of extending it over time for our needs. An advantage of this API is that it returns a Future. This would allow us at a later point to support the `async/await` style in JS if we want to. 

One current disadvantage of it is that it doesn't support timeouts. This could easily be supported as the Apache Exec has the notion of watchdogs that could do the work.


ThuF (approved):
+1 Seems good to me. @delchev do you have any comments, or we could merge this PR?

PR #1489, opened 2022-01-28
Configure variables for kyma and cloud foundry destination services

d-velev:
This PR makes it so that the environment variables needed for destination services in kyma and cloud foundry are set in dirigible with the same keys for both platforms. For kyma, the `DIRIGIBLE_OAUTH_PREFIX` and `DIRIGIBLE_DESTINATION_PREFIX` variables can be optionally set so that the service specific variables don't overlap.


d-velev:
@ThuF is there a better place for the `DIRIGIBLE_DESTINATION_` constants? There's an `OAuthService` class but no such class for the destination service.

releng/sap-cf-base/src/main/java/org/eclipse/dirigible/cf/utils/CloudFoundryUtils.java
```diff
@@ -165,6 +171,8 @@ public void setOrganizationName(String organizationName) {
 		@SerializedName("hana")
 		private List<HanaSchemaEnv> hanaSchemaEnv;
 
+		private List<DestinationEnv> destinationEnv;
```
dpanayotov, line 174: `@SerializedName("destination")`

releng/sap-kyma-base/src/main/java/org/eclipse/dirigible/kyma/KymaModule.java
```diff
@@ -77,6 +87,29 @@ private void configureOAuth() {
 		Configuration.setIfNull(OAuthService.DIRIGIBLE_OAUTH_APPLICATION_HOST, applicationHost);
 	}
 
+	private void configureDestination() {
+		String destinationPrefix = Configuration.get(DIRIGIBLE_DESTINATION_PREFIX, "");
```
dpanayotov, line 91: the default should be `destination_`, no?

dpanayotov (changes requested):

dpanayotov (approved):

Issue #1555, opened 2022-02-04
[XSOData] $skip parameter is causing an exception


dpanayotov:
Issue cloned from https://github.com/SAP/xsk/issues/1183

Executing a `GET` request to a XSOData endpoint with the `$skip` query parameter is causing an exception and no data is returned:
```
/services/v4/web/myapp/service.xsodata/Entity2?$skip=20&$top=20
```

However if `$skip=0`, then the expected data is returned and no exception is thrown:
```
/services/v4/web/myapp/service.xsodata/Entity1?$skip=0&$top=40
```

**Exception:**
```java
o.e.d.e.o.s.p.DefaultSQLProcessor - com.sap.db.jdbc.exceptions.SQLDataExceptionSapDB: Invalid argument resultSetType, use TYPE_FORWARD_ONLY.
org.apache.olingo.odata2.api.exception.ODataException: com.sap.db.jdbc.exceptions.SQLDataExceptionSapDB: Invalid argument resultSetType, use TYPE_FORWARD_ONLY.
	at org.eclipse.dirigible.engine.odata2.sql.processor.AbstractSQLProcessor.readEntitySet(AbstractSQLProcessor.java:256) ~[dirigible-odata-core-6.1.21.jar:na]
	at org.apache.olingo.odata2.core.Dispatcher.dispatch(Dispatcher.java:77) ~[olingo-odata2-core-2.0.11.jar:2.0.11]
	at org.apache.olingo.odata2.core.ODataRequestHandler.handle(ODataRequestHandler.java:131) ~[olingo-odata2-core-2.0.11.jar:2.0.11]
	at org.apache.olingo.odata2.core.batch.BatchHandlerImpl.handleBatchPart(BatchHandlerImpl.java:74) ~[olingo-odata2-core-2.0.11.jar:2.0.11]
	at org.eclipse.dirigible.engine.odata2.sql.processor.AbstractSQLProcessor.executeBatch(AbstractSQLProcessor.java:664) ~[dirigible-odata-core-6.1.21.jar:na]
	at org.apache.olingo.odata2.core.Dispatcher.dispatch(Dispatcher.java:190) ~[olingo-odata2-core-2.0.11.jar:2.0.11]
	at org.apache.olingo.odata2.core.ODataRequestHandler.handle(ODataRequestHandler.java:131) ~[olingo-odata2-core-2.0.11.jar:2.0.11]
[...]

PR #1556, opened 2022-02-04
[OData] Improve skip/offset logic

MarinHadzhiev:
Fix for #1555 

Change log:
1. **$skip** is now implemented with OFFSET for all DBs
2. **$top** is now implemented with LIMIT for all DBs
3. remove requirement for support of ResultSet **TYPE_SCROLL_INSENSITIVE** by the DB drivers


mit-ko:
looks good 👍 
just a remark: the limit and offset are added in ASE 16.

mit-ko (approved):

Issue #1539, opened 2022-02-02
[IDE] Debugger - JS files debugging is weird


ThuF:
I've created a hello world ***.js** file with the following content:

```javascript
var response = require("http/v4/response");

response.println("Hello World!");
response.flush();
response.close();
```

After I switch to the Debugger perspective and select the file, I go through the following screens, if I hit the **Continue** button only:

![debugger-continue-1](https://user-images.githubusercontent.com/4092083/152117949-3ccb74bb-8955-49e2-8514-c1a7092a6d26.png)
![debugger-continue-2](https://user-images.githubusercontent.com/4092083/152117957-179f8b59-0753-4ab8-8378-f5fd155a1721.png)

Inside **"/http/v4/response.js"**
![debugger-continue-3](https://user-images.githubusercontent.com/4092083/152117959-2173afc6-9629-480d-ab64-abb7f33d62a1.png)
![debugger-continue-4](https://user-images.githubusercontent.com/4092083/152117963-da3f7289-1fdd-4a92-ab62-7605fe36abfb.png)

Inside **"/io/v4/streams.js"**
![debugger-continue-5](https://user-images.githubusercontent.com/4092083/152117965-7a3786dc-4b22-40a0-8222-301842cbb53c.png)

Inside **"/io/v4/bytes.js"**
![debugger-continue-6](https://user-images.githubusercontent.com/4092083/152117968-af2142b1-a235-4acd-a7dd-2b2ed2cb2631.png)

It seems that a **debugger** is added to all required/imported modules.

It's also strange why my code is inside a **__wrap__** function:

```javascript
var __wrap__ = function (exports, require, module, __filename, __dirname) { 
debugger;

var response = require("http/v4/response");

response.println("Hello World!");
response.flush();
response.close();
}; __wrap__;
```

vmutafov:
@ThuF this should be closed now

ThuF:
Fixed with:
- https://github.com/eclipse/dirigible/pull/1567

PR #1567, opened 2022-02-08
Remove Module.debug.js module

tdermendjiev:
There are many logs and debugger statements in Module.debug.js which cause weird behavior when debugging. Keeping an both files may also cause problems if they diverge in future.
Related to https://github.com/eclipse/dirigible/issues/1539


ThuF (approved):

ThuF:
@tdermendjiev does this PR resolve https://github.com/eclipse/dirigible/issues/1539?

tdermendjiev:
@ThuF not completely, because the wrapper function is still there.

ThuF:
@tdermendjiev ok, I’m merging the PR, but not closing the issue.

PR #1570, opened 2022-02-08
Removing sensitive data from log

SAtanasovv:
### What does this PR do?
Hides sensitive data from being logged

- https://github.com/eclipse/dirigible/issues/1571


modules/services/service-core/src/main/java/org/eclipse/dirigible/runtime/core/initializer/DirigibleInitializer.java
```diff
@@ -88,6 +89,10 @@ private void printAllConfigurations() {
 		String[] keys = Configuration.getKeys();
 		for (String key : keys) {
 			String value = Configuration.get(key);
+			if(StringUtils.containsAnyIgnoreCase(key, "password", "secret"))
```
ThuF, line 92: Could you also include `user`?

ThuF (changes requested):
One minor comment

ThuF (approved):

PR #1572, opened 2022-02-09
Javascript http client option to put params in body only

dpanayotov:
Allow setting params only in body instead of in path when making requests 

related to https://github.com/SAP/xsk/issues/1228

### DIFF
--- a/api/api-javascript/api-http/src/main/resources/META-INF/dirigible/http/extensions/http.d.ts
+++ b/api/api-javascript/api-http/src/main/resources/META-INF/dirigible/http/extensions/http.d.ts
@@ -1232,5 +1232,9 @@ declare module "@dirigible/http" {
          * The SSL trust all enabled parameter
          */
         sslTrustAllEnabled?: boolean;
+        /**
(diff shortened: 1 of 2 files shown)


ThuF (changes requested):
I don't quite get this change isn't this enough: https://www.dirigible.io/api/http/client/#httpoptions

```javascript
httpClient.post("https://service.test.com/api/v1/", {
  text: {
    param1: "value1",
    param2: "value2",
  }
});
```

dpanayotov:
I looked at the options but none of them allow you to provide body with content type form-urlencoded

```javascript
    const jwtTokenResponse = httpClient.post(jwtTokenServiceUrl, {
        text: JSON.stringify({
            "username": username,
            "password": password
        }),
        headers: [
            {
                name: "Content-Type",
                value: "application/x-www-form-urlencoded",
            },
        ],
    });
```
results in text body with `Content-Type: text/plain`

The server does not expect a text body in json format, but a key-value pair as per https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.3

PR #1658, opened 2022-03-18
Fix permissions in documents perspective

### LINKED ISSUE #1064: [IDE] Documents - Constraint READ permission not working
**Describe the bug**
When defining a constraint with READ permission, then users with the corresponding role still are able to change documents (Upload new documents, delete and change). I would expect that if a user has only READ permission, that he then indeed also only has READ authorization. Further I would expect that then also no "Change options" (e.g. the delete button) is shown in the user interface.

**To Reproduce**
1. Go to Constraint configuration for a document service
2. Create a new constraint for a path, a role and permission READ
3. Then assign yourself this role and visit the document service URL
4. You will see that the READ permission does not work, instead you have full access

**Expected behavior**
only READ authorization is available if a READ constraint is defined

**Additional context**
We are using the SAP CMS version to handle our document store on SCP NEO. We have several constraints defined to handle access control based on maintained user roles.

tdermendjiev:
Fixes #1064


api/api-javascript/api-cms/src/main/resources/META-INF/dirigible/cms/cmis.js
```diff
@@ -1,13 +1,12 @@
...
  * are made available under the terms of the Eclipse Public License v2.0
  * which accompanies this distribution, and is available at
  * http://www.eclipse.org/legal/epl-v20.html
  *
- * SPDX-FileCopyrightText: 2022 SAP SE or an SAP affiliate company and Eclipse Dirigible contributors
- * SPDX-License-Identifier: EPL-2.0
+ * Contributors:
+ *   SAP - initial API and implementation
  */
 var cmis = require('cms/v4/cmis');
```
ThuF, line 11: Not intended change in the license headers

releng/sap-all-ephemeral/pom.xml
```diff
@@ -84,14 +84,35 @@
 		 	<groupId>org.eclipse.dirigible</groupId>
```
ThuF, line 84: Could you apply the same changes also for the:
- releng/sap-all
- releng/sap-cms

All `Neo` related modules are:
- releng/sap-all
- releng/sap-all-ephemeral
- releng/sap-cms

ThuF (changes requested):
As it seems there was an unintended change of the license headers. To fix them do the following:

1. Run `mvn clean install -P content` to fetch the changes coming from DirigibleLabs (note: this will mess the license headers as well).
2. Run `mvn clean install` to fix/update the license headers.

As you've already performed Step 1. could you try just to run `mvn clean install` from the root folder and to check whether the license headers are updated as expected?

**Update:** Also could you list the issues that are being fixed with this PR?

PR #1682, opened 2022-03-31
fix search view

ivanvolkoff:
Fix non-working search view.


StanZGenchev:
There are two problems with this PR:

1. Changes are not reflected in the ide-workspace repo, so it cannot be merged.
2. It does not fix the issue with is either a back-end issue or an issue in search.js

ivanvolkoff:
@StanZGenchev there's no need to add changes to the ide-workspace menu, pr represents just an update of ide-workspace with -Pcontent command, and its solves the search issue.

PR #1709, opened 2022-04-17
OData now uses row_num value as ID for HANA views

BorisNen:
Related to https://github.com/SAP/xsk/issues/1444

Changelog:
1. ODataEntityDefinition now has a separate field for generated keys.
2. OData2ODataMTransformer now checks the keyGenerated field and adds it to the json.
3. SQLSelectClause will add HANA's row_num value to the select clause when we have keyGenerated in the table binding.
4. AbstractSQLProcessor now builds two different ResultSetEntities - one that uses the table keys for entities and one that uses the row_num values as key when we have a HANA view.

### DIFF
--- a/modules/odata/odata-core/src/main/java/org/eclipse/dirigible/engine/odata2/sql/processor/ResultSetReader.java
+++ b/modules/odata/odata-core/src/main/java/org/eclipse/dirigible/engine/odata2/sql/processor/ResultSetReader.java
@@ -41,27 +41,14 @@ public class ResultSetReader {
         return result;
     }
 
-
-    protected ResultSetEntity getResultSetEntityForView(SQLSelectBuilder selectEntityQuery, final EdmEntityType entityType,
-                                                        Collection<EdmProperty> properties, final ResultSet resultSet) throws SQLException, ODataException, IOException {
-
-        return getResultSetEntityForObject(selectEntityQuery, entityType, properties, resultSet, true);
-    }
-
     protected ResultSetEntity getResultSetEntity(SQLSelectBuilder selectEntityQuery, final EdmEntityType entityType,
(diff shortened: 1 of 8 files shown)


modules/odata/odata-core/src/main/java/org/eclipse/dirigible/engine/odata2/sql/processor/ResultSetReader.java
```diff
@@ -42,12 +42,28 @@ protected Map<String, Object> getEntityDataFromResultSet(SQLSelectBuilder select
     }
 
 
+    protected ResultSetEntity getResultSetEntityForView(SQLSelectBuilder selectEntityQuery, final EdmEntityType entityType,
```
MarinHadzhiev, line 45: ForView is misleading, because yo can have generated key also for table. May be it will be better to remove getResultSetEntityForView and getResultSetEntity and rename getResultSetEntityForObject to get ResultSetEntity with last input parameter renamed from isView to hasGeneratedId. And just remove the if statement from AbstractSQLProcessor to use this new method.

ThuF (commented):
+1 seems good to me, the comments from @MarinHadzhiev should be addressed first, before merging the PR.

modules/odata/odata-core/src/main/java/org/eclipse/dirigible/engine/odata2/sql/processor/ResultSetReader.java
BorisNen: Fixed in [b6ad0d1](https://github.com/eclipse/dirigible/pull/1709/commits/b6ad0d13aca48d01ba9430b4a4b0ebab3778d71c)

MarinHadzhiev (approved):

PR #1748, opened 2022-05-13
Odata with HANA calculation views

aihanrashidov:
Resolves https://github.com/SAP/xsk/issues/1499

Modified the `OData2ODataXTransformer` to add keys and properties from the odata definition's parameters if there are such. Also modified the `OData2ODataMTransformer` to add `dataStructureType` and `_parameters_` bindings. 

Modified the odata core module to build proper SQL when reading HANA calculation view entities with input parameters.


modules/odata/odata-core/src/main/java/org/eclipse/dirigible/engine/odata2/sql/binding/EdmTableBinding.java
```diff
@@ -217,6 +217,14 @@ public ColumnInfo getColumnInfo(EdmProperty property) {
         }
     }
 
+    public List<String> getParameters() throws EdmException {
+        return readMandatoryConfig("_parameters_", List.class);
+    }
+
+    public String getDataStructureType() throws EdmException {
+        return readMandatoryConfig("dataStructureType", String.class);
+    }
```
MarinHadzhiev, line 226: Let's make them not mandatory but optional with default values in case there are missing. Otherwise we will break current functionality and working applications.

modules/odata/odata-core/src/main/java/org/eclipse/dirigible/engine/odata2/sql/clause/SQLSelectClause.java
```diff
@@ -192,46 +232,75 @@ private String buildLimit(final SQLContext context) {
...
+            }
+
+            if (pi.hasNext()) {
+                parameters.append(",");
+            }
+        }
+
+        return "(" + parameters + ")";
+    }
+
```
MarinHadzhiev, line 279: What if DB is not HANA? This will return () can we move the check from line 268 up so that it skips the whole method execution if DB type is not HANA?

MarinHadzhiev (changes requested):

MarinHadzhiev (approved):

PR #1898, opened 2022-06-27
OData CUD procedure handlers

cheezus1:
AbstractSQLProcessor now checks if there is an error document present in the context, generated during handler execution.


modules/odata/odata-core/src/main/java/org/eclipse/dirigible/engine/odata2/sql/processor/AbstractSQLProcessor.java
```diff
@@ -330,11 +331,17 @@ public ODataResponse createEntity(final PostUriInfo uriInfo, final InputStream c
             updateCreateEventHandlerContext(handlerContext, connection, uriInfo, entry);
             this.odata2EventHandler.beforeCreateEntity(uriInfo,
                     requestContentType, contentType, entry, handlerContext);
+            if(handlerContext.containsKey(ERROR_DOCUMENT) && handlerContext.get(ERROR_DOCUMENT) != null) {
```
BorisNen, line 334: Could you move the conditions to a separate method called isErrorDocumentPresent() and then use the method if(isErrorDocumentPresent()) -> return get(ERROR_DOCUMENT) for all checks below?

BorisNen (changes requested):
Could you move the conditions to a separate method called isErrorDocumentPresent() and then use the method if(isErrorDocumentPresent()) -> return get(ERROR_DOCUMENT) for all checks below?

BorisNen (approved):

modules/odata/odata-core/src/main/java/org/eclipse/dirigible/engine/odata2/sql/processor/AbstractSQLProcessor.java
```diff
@@ -330,11 +331,17 @@ public ODataResponse createEntity(final PostUriInfo uriInfo, final InputStream c
             updateCreateEventHandlerContext(handlerContext, connection, uriInfo, entry);
             this.odata2EventHandler.beforeCreateEntity(uriInfo,
                     requestContentType, contentType, entry, handlerContext);
+            if (isErrorDocumentPresent(handlerContext)) {
+                return (ODataResponse) handlerContext.get(ERROR_DOCUMENT);
+            }
```
MarinHadzhiev, line 336: getting the result of executing method from input param of the method is bad. Why not just to return ODataResponse from the handler method?

modules/odata/odata-core/src/main/java/org/eclipse/dirigible/engine/odata2/sql/processor/AbstractSQLProcessor.java
```diff
@@ -407,6 +417,9 @@ public ODataResponse createEntity(final PostUriInfo uriInfo, final InputStream c
         try (Connection connection = getDataSource().getConnection()) {
             updateCreateEventHandlerContext(handlerContext, connection, uriInfo, entry);
             this.odata2EventHandler.afterCreateEntity(uriInfo, requestContentType, contentType, entry, handlerContext);
+            if (isErrorDocumentPresent(handlerContext)) {
+                return (ODataResponse) handlerContext.get(ERROR_DOCUMENT);
+            }
```
MarinHadzhiev, line 422: same as above comment

modules/odata/odata-core/src/main/java/org/eclipse/dirigible/engine/odata2/sql/processor/AbstractSQLProcessor.java
```diff
@@ -389,6 +396,9 @@ public ODataResponse createEntity(final PostUriInfo uriInfo, final InputStream c
                     }
                     updateCreateEventHandlerContext(handlerContext, connection, uriInfo, entry);
                     this.odata2EventHandler.afterCreateEntity(uriInfo, requestContentType, contentType, entry, handlerContext);
+                    if (isErrorDocumentPresent(handlerContext)) {
+                        return (ODataResponse) handlerContext.get(ERROR_DOCUMENT);
+                    }
```
MarinHadzhiev, line 401: same as above comment

MarinHadzhiev (changes requested):
in general getting the result of method execution from an input param of the method is bad. Why not just to return ODataResponse from the handler method?

modules/odata/odata-core/src/main/java/org/eclipse/dirigible/engine/odata2/sql/processor/AbstractSQLProcessor.java
```diff
@@ -325,17 +327,29 @@ public ODataResponse createEntity(final PostUriInfo uriInfo, final InputStream c
...
                         requestContentType, contentType, handlerContext);
-                entryMap = (HashMap<String, Object>) handlerContext.get(ENTRY_MAP);
+                if (isErroneousResponse(response)) {
+                    return response;
+                } else if (response != null) {
+                    ByteArrayOutputStream entityOutputStream = new ByteArrayOutputStream();
+                    ((InputStream) response.getEntity()).transferTo(entityOutputStream);
+                    InputStream entityInputStream = new ByteArrayInputStream(entityOutputStream.toByteArray());
+                    InputStream entityCloneInputStream = new ByteArrayInputStream(entityOutputStream.toByteArray());
+                    entry = parseEntry(entitySet, entityInputStream, response.getContentHeader(), false);
```
MarinHadzhiev, line 350: Do you need this logic? Why not just return the response or ODataResponse.fromResponse(response)? And all streams must be closed.

modules/odata/odata-core/src/main/java/org/eclipse/dirigible/engine/odata2/sql/processor/AbstractSQLProcessor.java
```diff
@@ -362,13 +376,15 @@ public ODataResponse createEntity(final PostUriInfo uriInfo, final InputStream c
...
 
-                ODataResponse response;
                 try (Connection connection = getDataSource().getConnection()) {
-                    if (entryMap != null) {
-                        response = ExpandCallBack.writeEntryWithExpand(getContext(), //
-                                (UriInfo) uriInfo, //
-                                entryMap, //
-                                contentType);
+                    updateCreateEventHandlerContext(handlerContext, connection, uriInfo, entry);
+                    ODataResponse afterCreateEntityResponse = this.odata2EventHandler.afterCreateEntity(uriInfo, requestContentType, contentType, entry, handlerContext);
```
MarinHadzhiev, line 381: if you call afterCreateEntity here you should not call it again after the if statement bellow, actually you can't reach the code bellow

modules/odata/odata-core/src/main/java/org/eclipse/dirigible/engine/odata2/sql/processor/AbstractSQLProcessor.java
```diff
@@ -398,15 +414,18 @@ public ODataResponse createEntity(final PostUriInfo uriInfo, final InputStream c
...
+            response = this.odata2EventHandler.afterCreateEntity(uriInfo, requestContentType, contentType, entry, handlerContext);
+            if (isErroneousResponse(response)) {
+                return response;
+            }
+
+            ODataContext context = getContext();
+            EntityProviderWriteProperties writeProperties = EntityProviderWriteProperties
+                    .serviceRoot(context.getPathInfo().getServiceRoot()).expandSelectTree(entry.getExpandSelectTree())
+                    .build();
+            return EntityProvider.writeEntry(contentType, entitySet, entry.getProperties(), writeProperties);
```
MarinHadzhiev, line 428: Do you need this logic? It the same as above - copy/paste?

MarinHadzhiev (changes requested):

modules/engines/engine-odata/src/main/java/org/eclipse/dirigible/engine/odata2/handler/ScriptingOData2EventHandler.java
```diff
@@ -39,346 +39,351 @@
...
-			context.put("contentType", contentType);
-			context.put("entry", entry);
-			executeHandlers(handlers, context);
-		} catch (EdmException | ODataException e) {
-			logger.error(e.getMessage(), e);
-		} catch (ScriptingException e) {
-			throw new ODataException(ERROR_EXECUTING_XSJSLIB_HANDLER + e.getMessage(), e);
-		}
-	}
+    private static final String ERROR_EXECUTING_XSJSLIB_HANDLER = "Error executing .xsjslib handler: ";
```
MarinHadzhiev, line 50: This is generic scripting handler for the dirigible, it is not only for xsjslib functions. It is also used with the standard Dirigible OData events where the handler function could be in a js file. This should be a generic message.

modules/odata/odata-core/src/main/java/org/eclipse/dirigible/engine/odata2/sql/processor/AbstractSQLProcessor.java
```diff
@@ -634,25 +657,70 @@ public BatchResponsePart doExecuteChangeSet(final BatchHandler handler, final Li
...
+    private boolean isErroneousResponse(ODataResponse response) {
+        // erroneous requests are >= 400 (BAD_REQUEST)
+        return response != null && response.getStatus().getStatusCode() >= HttpStatusCodes.BAD_REQUEST.getStatusCode();
+    }
+
+    private void checkForKeys(ODataEntry entry, List<EdmProperty> keyProperties) throws ODataException {
+        if (!keyProperties.isEmpty()) {
+            for (EdmProperty keyProperty : keyProperties) {
+                if (entry.getProperties().get(keyProperty.getName()) == null) {
+                    throw new ODataException("Cannot create entity without key(s)");
```
MarinHadzhiev, line 669: add info about the entry like name or id to the exception message, it will be easier to debug it.

modules/odata/odata-core/src/main/java/org/eclipse/dirigible/engine/odata2/sql/processor/AbstractSQLProcessor.java
```diff
@@ -634,25 +657,70 @@ public BatchResponsePart doExecuteChangeSet(final BatchHandler handler, final Li
...
+    }
+
+    private void putEntryInsertBuilderInContext(Map<Object, Object> context, String key, UriInfo uriInfo, EdmEntitySet entitySet) throws ODataException {
+        if (context.containsKey(key)) {
+            String beforeUpdateEntryJSON = (String) context.get(key);
+            InputStream entryContent = new ByteArrayInputStream(beforeUpdateEntryJSON.getBytes(StandardCharsets.UTF_8));
+            ODataEntry parsedEntry = parseEntry(entitySet, entryContent, ContentType.APPLICATION_JSON.toContentTypeString(), false);
+            context.put(INSERT_BUILDER_CONTEXT_KEY, this.getSQLQueryBuilder().buildInsertEntityQuery(uriInfo, parsedEntry, getContext()));
+        }
     }
```
MarinHadzhiev, line 725: Is it possible to put in the context - uriInfo, entry, getDataSource and ODataContext everything else could be build based on them. This way you will not need to create and add sql queries specific for the ScriptingHandler implementation and each implementation will be able to use them to implement it's logic.

Also the input stream should be closed.

MarinHadzhiev (changes requested):

MarinHadzhiev (approved):

PR #2077, opened 2022-09-28
Data structures name encapsulation fix

aihanrashidov:
When creating a data structure in dirigible with the `DIRIGIBLE_DATABASE_NAMES_CASE_SENSITIVE` configuration set to `true` the name is being encapsulated within quotes. However there is an issue for data structure names with this sample convention `test::db.users` which is seen in applications using the `HANA` database. The encapsulate method would currently split this into `"test::db"."users"` which is incorrect. This PR fixes the names to be properly encapsulated.


ThuF (approved):

ThuF:
+1 and good to me

PR #2189, opened 2022-11-02
Spring jobs

SAtanasovv:
Added Spring Boot for JobLog ,JobEmail and JobParameter


components/core-jobs/pom.xml
```diff
@@ -36,7 +26,31 @@
             <version>7.0.0-SNAPSHOT</version>
             <scope>compile</scope>
         </dependency>
-    </dependencies>
+        <dependency>
+            <groupId>org.eclipse.dirigible</groupId>
+            <artifactId>dirigible-engine-api</artifactId>
```
delchev, line 31: why this module is used?

components/core-jobs/src/main/java/org/eclipse/dirigible/components/jobs/synchronizer/JobSynchronizer.java
```diff
@@ -173,6 +186,19 @@ public void cleanup(Job artefact) {
...
     @Override
     public boolean complete(TopologyWrapper<Artefact> wrapper, String flow) {
+        Job job = null;
+        if (wrapper.getArtefact() instanceof Job){
+            job = (Job) wrapper.getArtefact();
+            Job existingJob = jobService.findByName(job.getName());
+            List<JobEmail> emailDefinitions = jobEmailService.findAllByName(existingJob.getName());
+
+            jobLogService.jobLogged(existingJob.getName(), existingJob.getHandler(), existingJob.getMessage());
+            JobEmailProcessor.createAndSendJobEmail(job, existingJob, emailDefinitions);
```
delchev, line 196: does not seem to be correct to use processor here
it should just cover the LM of the Job artefact

delchev (changes requested):

PR #2450, opened 2023-03-22
Terminal config

krasimirdermendzhiev:
### What does this PR do?

Add terminal property - default `false`.

Add option for terminal.


ThuF:
@krasimirdermendzhiev not sure where we should set the default value to `true` or ` false`? Maybe we can enable it by default and turn it off via the environment variable?

PR #2547, opened 2023-05-29
Fixes keyboard shortcut behaviour in projects view

### LINKED ISSUE #2151: [Workspace] Inline Rename does not work
Inline Rename does not work in the Workspace tree.
When click F2 it opens inline text box which does not call the backend service.
When it opens via popup menu and the Rename Dialog it works.

It should be either disable inline editing or link to the backend.

StanZGenchev:
### What does this PR do?

Configures the 'F2/Rename', 'Enter/Double click' and Delete keyboard shortcuts in the Projects view.
Adds new "Ctrl+C" (Copy), "Ctrl+X" (Cut) and "Ctrl+V" (Paste) shortcuts.

Fixes #2151


ThuF (approved):
Seems good to me.

delchev (approved):

PR #2640, opened 2023-07-18
feat: add typescript support

vmutafov:
This PR adds initial TypeScript support. It could be tried out by, for example, creating the following TypeScript files in the IDE: 

```TS 
// lib.ts
export const a: number = 123;
```

```TS 
// service.ts
import { response } from "@dirigible/http";
import { a } from "./lib"

response.println(`a is ${a}`);
response.flush();
response.close();
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,7 @@ Nevertheless, we highly recommend building the binaries from source in order to 
 - [Git](http://git-scm.com/)
 - [Java JDK 11+](https://adoptopenjdk.net/)
 - [Maven 3.5.x](http://maven.apache.org/docs/3.5.3/release-notes.html)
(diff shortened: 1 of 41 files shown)


ThuF:
@vmutafov it seems that the compiled `*.mjs` files are placed in the root folder, instead of the corresponding sub-folders:

Screenshot 2023-07-22 at 17 27 07

This resulting into the following error message when accessing `/services/ts/university-ts-new/entities/student/index.ts`:

```
JavaScript source code for project name 'university-ts-new' and file name 'entities/student/index.mjs' could not be found, consider publishing it.
```

However accessing `/services/ts/university-ts-new/index.ts` results into successful execution.

PR #2649, opened 2023-07-27
feat: compile ts projects in registry on spring start

vmutafov:


components/engine-typescript/src/main/java/org/eclipse/dirigible/TypeScriptEndpoint.java
```diff
@@ -9,9 +9,10 @@
  * SPDX-FileCopyrightText: 2022 SAP SE or an SAP affiliate company and Eclipse Dirigible contributors
  * SPDX-License-Identifier: EPL-2.0
  */
-package org.eclipse.dirigible.components.engine.javascript.endpoint;
+package org.eclipse.dirigible;
```
delchev, line 12: is there a reason for shortening the package instead of e.g. org.eclipse.dirigible.components.engine.typescript

components/core-initializers/src/main/java/org/eclipse/dirigible/components/initializers/synchronizer/SynchronizationInitializer.java
```diff
@@ -53,6 +64,7 @@ public void handleContextStart(final ApplicationReadyEvent are) {
 		synchronizationProcessor.prepareSynchronizers();
 		classpathExpander.expandContent();
 		synchronizationProcessor.processSynchronizers();
+		initializers.forEach(Initializer::initialize);
```
delchev, line 67: maybe initialize() could throw runtime exception for one of the Initializers?

delchev (changes requested):

components/engine-typescript/src/main/java/org/eclipse/dirigible/TypeScriptEndpoint.java
vmutafov: No, I noticed it as well and will change it

components/core-initializers/src/main/java/org/eclipse/dirigible/components/initializers/synchronizer/SynchronizationInitializer.java
vmutafov: you’re right, I’ll update this

delchev (approved):

PR #2660, opened 2023-08-02
add .wasm content-type

Fluctuationqt:
- Adds `.wasm` content type.


vmutafov:
What editor would be used for webassembly files?

ThuF (approved):

PR #2662, opened 2023-08-03
[GH-#2661] Added missing setBinaryStream function

### LINKED ISSUE #2661: Add missing setBinaryStream in DB API
Add missing `setBinaryInput` for prepared statements in the DB API and update the documentation.

nedelcho-delchev-tues:
### What does this PR do?
Adds missing setBinaryStream function in DB API and updated the [documentation](https://www.dirigible.io/api/database/database/#preparedstatement).

fixes #2661

### DIFF
--- a/components/api-database/src/main/resources/META-INF/dirigible/db/database.js
+++ b/components/api-database/src/main/resources/META-INF/dirigible/db/database.js
@@ -206,6 +206,14 @@ function PreparedStatement(internalStatement) {
 		this.native.setNull(index, sqlType);
 	};
 
+	this.setBinaryStream = function (parameter, inputStream, length) {
+		if (length !== undefined && length !== null) {
+			this.native.setBinaryStream(parameter, inputStream, length);
+		} else {
+			this.native.setBinaryStream(parameter, inputStream);
+		}
+	};
+
 	this.setBoolean = function (index, value) {
 		if (value !== null && value !== undefined) {
 			this.native.setBoolean(index, value);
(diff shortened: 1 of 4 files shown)


components/api-database/src/main/resources/META-INF/dirigible/db/database.js
```diff
@@ -206,6 +206,14 @@ function PreparedStatement(internalStatement) {
 		this.native.setNull(index, sqlType);
 	};
 
+	this.setBinaryStream = function (parameter, inputStream, length) {
+		if (length !== undefined && length !== null) {
```
vmutafov, line 210: `if(length)` is enough here

components/api-database/src/main/resources/META-INF/dirigible/db/extensions/db.d.ts
```diff
@@ -335,6 +335,14 @@ declare module "@dirigible/db" {
          */
         setClob(index:number, value);
 
+        /**
+         * Sets a parameter of type binary strem
+         * @param index
+         * @param inputStream
+         * @param length length of the input stream. If not provided gets all the binary data
+         */
+        setBinaryStream(index: number, inputStream, length?)
```
vmutafov, line 344: We should add the type for the `length` parameter

vmutafov: And do we know the `inputStream`'s type?

vmutafov (changes requested):

ThuF (approved):
+1 Seems good to me.

PR #2785, opened 2023-09-26
chore: unpack spring boot executable jar before run and update docs

vmutafov:


ThuF:
@vmutafov I guess this would affect how "Custom Stacks" are created, for example:
- https://github.com/codbex/codbex-kronos/blob/main/application/Dockerfile
- https://github.com/codbex/codbex-atlas/blob/main/application/Dockerfile
- https://github.com/codbex/codbex-helios/blob/main/application/Dockerfile

vmutafov:
@ThuF yes, you're right, but I plan to add a `Dockerfile.base` that would create a base image where we have all our dependency tools like `esbuild`, `tsc`, etc., and have this base image used where necessary. I'll convert this PR to a draft one for now

PR #2949, opened 2023-11-09
add formatting doc for intellij and vscode

iliyan-velichkov:


vmutafov (approved):
I've added a step for IntelliJ to use the formatter config in the project instead of globally. Works fine from what I've tried

ThuF (approved):
It seems that the formatter also works well for **VSCode**.

PR #2958, opened 2023-11-09
fix: decorators module rename and default exports, fix rs arrow function args

vmutafov:
The following should work in both ESM and TS files:
``` TypeScript
import { Controller, Get, Post } from "@dirigible/http"
// or if you use the CJS modules support
// const { Controller, Get, Post } = require("http/v4/rs/decorators");

@Controller
class MyApi {

    @Get("/")
    onGet() {
        return "Hello there!";
    }

    @Post("/")
    onPost(body) {
        return {
            some: "data"
        }
    }
}
```


ThuF (approved):
+1 Seems good to me.

PR #2970, opened 2023-11-14
chore: format on mvn install and validate as extra step in gh action

vmutafov:


delchev (approved):

pom.xml
```diff
@@ -110,10 +110,10 @@
 						</configuration>
 				        <executions>
 							<execution>
-							<id>validate-java-code-formatting</id>
-							<phase>validate</phase>
+							<id>format-java-code</id>
+							<phase>install</phase>
```
iliyan-velichkov, line 114: better format the code before the compilation and tests execution for just in case

iliyan-velichkov (changes requested):

pom.xml
vmutafov: Fixed. Will do this in the `process-sources` phase

iliyan-velichkov: better use `validate`

vmutafov: It's not good to use `validate` for such cases. If there is any source generation, `validate` would be too early to format

ThuF (approved):
+1 Seems good to me.

delchev (approved):

PR #2981, opened 2023-11-15
fix master formatting

iliyan-velichkov:


iliyan-velichkov:
@delchev @ThuF @vmutafov

Issue #2982, opened 2023-11-15
[IDE] Editor is closed on file move or rename


StanZGenchev:
**Describe the bug**
When you open a file in an editor and then rename the file, the editor is closed and you have to open it again manually.
When you open a file in an editor and then move the file, the editor is not closed but it will not save the file. This can lead to data loss.

**Expected behavior**
When a user moves or renames a file that is opened in an editor, the editor should get the new path and continue to work without the need to reload or close it.

**Desktop:**
 - OS: Fedora Linux 38
 - Browser: Firefox, Chrome
 - Version: Dirigible 8.18.6

PR #2983, opened 2023-11-15
[IDE] #2982 - Editor is closed on file move or rename

StanZGenchev:
### What does this PR do?

Fixes the file move and rename behavior when the file is opened in an editor.

#2982


ThuF (approved):
+1 Seems good to me.

Issue #3005, opened 2023-11-20
[DSM] Move all buttons from the bottom toolbar to the top toolbar


StanZGenchev:
**Describe the issue**
This issue is identical to #2876 but for the DSM editor.

**Desktop:**
 - OS: Fedora Linux 38
 - Browser: Any
 - Version: Dirigible 8.18.6

PR #3006, opened 2023-11-20
[DSM] #3005 - Move all buttons from the bottom toolbar to the top toolbar

StanZGenchev:
### What does this PR do?

Moves all buttons from the bottom toolbar to the top toolbar in the DSM/Schema editor.

#3005


ThuF (approved):
+1 Seems good to me.

PR #3032, opened 2023-11-23
fix devtools version

iliyan-velichkov:
fixes actuator endpoint


iliyan-velichkov:
@ThuF @delchev @vmutafov review it pls

PR #3068, opened 2023-12-05
Dependencies from springboot

iliyan-velichkov:
- use spring managed dependencies to benefit from the compatibility
- add dirigible dependencies as managed to
  - remove versions declarations
  - enable using dirigible parent as parent of your project
  - using dirigible parent as pom import
- remove unused properties

### DIFF
--- a/build/application/pom.xml
+++ b/build/application/pom.xml
@@ -37,14 +37,12 @@
 		<dependency>
 			<groupId>org.eclipse.dirigible</groupId>
 			<artifactId>dirigible-commons-config</artifactId>
-			<version>9.0.0-SNAPSHOT</version>
 		</dependency>
 
 		<!-- Spring Boot -->
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-web</artifactId>
-			<version>${spring.boot.version}</version>
 			<exclusions>
 				<exclusion>
 					<groupId>org.apache.logging.log4j</groupId>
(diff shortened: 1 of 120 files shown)


iliyan-velichkov:
@ThuF @vmutafov @delchev the PR is for review

ThuF:
@iliyan-velichkov could you double check if setting a new version (e.g. as in the Release action) works as expected? Also could you check if the correct version is displayed in the `About` view?

iliyan-velichkov:
> @iliyan-velichkov could you double check if setting a new version (e.g. as in the Release action) works as expected? Also could you check if the correct version is displayed in the `About` view?

here is the "About", I think it is fine
image

iliyan-velichkov:
> @iliyan-velichkov could you double check if setting a new version (e.g. as in the Release action) works as expected? Also could you check if the correct version is displayed in the `About` view?

`mvn versions:set -DnewVersion=10.0.0` works as well

iliyan-velichkov:
> It would be hard to maintain the correct list of all Dirigible dependencies in the main `pom.xml`.
> 
> I guess it would be better to leave this improvement for another time, after we introduce "groups" of maven dependencies e.g. `dirigible-components`.
> 
> Also could you create an issue about it and an example, how this improvement can be used in a custom dirigible stacks. Something similar to https://www.dirigible.io/help/tutorials/customizations/custom-stack/ (note that this is tutorial is working with version 6.x).

regarding the documentation, I will create such an example once the PR is merged

@ThuF why do you think that maintenance is tough? I even think that this way, the management is easier. You just put the dependencies which are need in the `dependencyManagement`. I think the pros are more than cons + this is a standard approach for dependencies management. This is what spring boot does.
Having in mind that the PR is ready, I think it is better to merge it and refactor it once these groups are implemented.
+ It will be easier to create a dirigible custom stack and the approach will be the same as you should use with Spring Boot.

iliyan-velichkov:
created a separate PR - https://github.com/eclipse/dirigible/pull/3071

PR #3081, opened 2023-12-07
Add azure authz support

mvmanov:
### What does this PR do?
Add azure authz support

### DIFF
--- a/build/application/pom.xml
+++ b/build/application/pom.xml
@@ -171,7 +171,11 @@
 			<artifactId>dirigible-components-security-oauth2</artifactId>
 			<version>9.0.0-SNAPSHOT</version>
 		</dependency>
-
+		<dependency>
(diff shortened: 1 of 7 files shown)


iliyan-velichkov:
@delchev is this PR still relevant?

delchev:
Thanks for the contribution, and apologies for the long silence on it. Closing as superseded: the platform now has a shared OIDC layer in `security-oauth2` (`TenantAwareAuthoritiesMapper` over a configurable groups claim, tenant-group resolution, session revalidation, native login, RP-initiated logout) that the Keycloak and Cognito profiles build on as thin `@Profile` configurations, and an Entra ID profile is best added the same way rather than as a separate filter chain with a static group-to-role map. The PR also targets 9.0.0-SNAPSHOT and conflicts with master. The follow-up is tracked in #7478.

Issue #3078, opened 2023-12-06
[IDE] Database - Can't open the SystemDB


ThuF:
Selecting the `SystemDB` from the **Database Perspective** fails with the following error:

```java
2023-12-06 18:22:57.347 [INFO ] [http-nio-8080-exec-10] o.e.d.c.d.s.m.DataSourcesManager - Initializing a datasource with name: SystemDB
2023-12-06 18:22:57.354 [INFO ] [http-nio-8080-exec-10] com.zaxxer.hikari.HikariDataSource - SystemDB - Starting...
2023-12-06 18:22:57.358 [INFO ] [http-nio-8080-exec-10] com.zaxxer.hikari.HikariDataSource - SystemDB - Start completed.
2023-12-06 18:22:57.362 [ERROR] [http-nio-8080-exec-10] o.a.c.c.C.[.[.[.[dispatcherServlet] - Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception [Request processing failed; nested exception is java.lang.IllegalStateException: Could not register object [org.eclipse.dirigible.components.data.sources.manager.ManagedDataSource@59a0dbdf] under bean name 'SystemDB': there is already object [HikariDataSource (HikariPool-1)] bound] with root cause
java.lang.IllegalStateException: Could not register object [org.eclipse.dirigible.components.data.sources.manager.ManagedDataSource@59a0dbdf] under bean name 'SystemDB': there is already object [HikariDataSource (HikariPool-1)] bound
        at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.registerSingleton(DefaultSingletonBeanRegistry.java:124) ~[spring-beans-5.3.31.jar!/:5.3.31]
        at org.springframework.beans.factory.support.DefaultListableBeanFactory.registerSingleton(DefaultListableBeanFactory.java:1149) ~[spring-beans-5.3.31.jar!/:5.3.31]
        at org.eclipse.dirigible.components.data.sources.manager.DataSourcesManager.initializeDataSource(DataSourcesManager.java:166) ~[dirigible-components-data-sources-9.0.0-SNAPSHOT.jar!/:na]
        at org.eclipse.dirigible.components.data.sources.manager.DataSourcesManager.getDataSource(DataSourcesManager.java:104) ~[dirigible-components-data-sources-9.0.0-SNAPSHOT.jar!/:na]
[...]

PR #3085, opened 2023-12-07
fix: do not register systemdb bean a second time

vmutafov:
Fixes: https://github.com/eclipse/dirigible/issues/3078


delchev (approved):

ThuF (approved):
+1 seems good to me

PR #3119, opened 2023-12-20
feat: Add integration tests for Camel PlatformHTTP and @dirigible/integrations API

Fluctuationqt:
- Modified JS invoker to return response to any invoking camel routes.
- Adds an integration test which calls a PlatformHttp route, which in turn provides the payload to a dirigible js script that forms a response.
- Adds an integration test which from a dirigible js script calls a camel direct route via the js api that forms a response.
- Dirigible code runner closes context after route completes when camel invokes dirigible js/mjs/ts files

### DIFF
--- a/components/api/api-modules-javascript/src/main/resources/META-INF/dirigible/modules/src/integrations/index.ts
+++ b/components/api/api-modules-javascript/src/main/resources/META-INF/dirigible/modules/src/integrations/index.ts
@@ -1,7 +1,7 @@
 const SpringBeanProvider = Java.type("org.eclipse.dirigible.components.spring.SpringBeanProvider");
 const Invoker = Java.type('org.eclipse.dirigible.components.engine.camel.invoke.Invoker');
 const invoker = SpringBeanProvider.getBean(Invoker.class);
-const String = Java.type('java.lang.String');
+const CamelMessage = Java.type('org.apache.camel.Message');
 
 export function invokeRoute(routeId, payload, headers) {
     return invoker.invokeRoute(routeId, payload, headers);
@@ -11,53 +11,27 @@ export function getInvokingRouteMessage() {
     return __context.camelMessage;
 }
 
-// Define an interface for the headers
-export interface Headers {
-    [key: string]: any;
+export interface HeadersMap {
+    [key: string]: string | string[];
 }
 
-// Wrapper class for Camel Message
-export class IntegrationMessage {
-    private message: any; // Replace 'any' with the actual type of Camel Message
+export interface IntegrationMessage {
 
-    constructor(message: any) {
-        this.message = message;
-    }
+    constructor(message: any);
 
-    // Get the body of the message
-    getBody(): String {
-        return this.message.getBody(String.class);
-    }
+    getBody(): any;
 
-    // Set the body of the message
-    setBody(body: String): void {
-        this.message.setBody(body);
-    }
+    getBodyAsString(): string;
 
-    // Get headers of the message
-    getHeaders(): Headers {
-        const headers: Headers = {};
-        const headerNames = this.message.getHeaders().keySet().toArray();
-        for (const headerName of headerNames) {
-            headers[headerName] = this.message.getHeader(headerName);
-        }
-        return headers;
-    }
+    setBody(body: any): void;
 
-    // Set headers of the message
-    setHeaders(headers: Headers): void {
-        for (const [key, value] of Object.entries(headers)) {
-            this.message.setHeader(key, value);
-        }
-    }
+    getHeaders(): HeadersMap;
 
-    // Set header of the message
-    setHeader(key: String, value: any): void {
-        this.message.setHeader(key, value);
-    }
+    getHeader(key: string): string | string[];
 
-    // Returns the original camel message
-    getCamelMessage(): any {
-        return this.message;
-    }
+    setHeaders(headers: HeadersMap): void;
+
+    setHeader(key: string, value: string | string[]): void;
+
+    getCamelMessage(): typeof CamelMessage;
 }
--- a/components/api/api-modules-javascript/src/main/resources/META-INF/dirigible/modules/src/integrations/index.ts
+++ b/components/api/api-modules-javascript/src/main/resources/META-INF/dirigible/modules/src/integrations/index.ts
@@ -10,3 +10,28 @@ export function invokeRoute(routeId, payload, headers) {
 export function getInvokingRouteMessage() {
     return __context.camelMessage;
 }
+
+export interface HeadersMap {
+    [key: string]: string | string[];
+}
+
+export interface IntegrationMessage {
+
+    constructor(message: any);
+
+    getBody(): any;
+
+    getBodyAsString(): string;
+
+    setBody(body: any): void;
+
+    getHeaders(): HeadersMap;
+
+    getHeader(key: string): string | string[];
+
+    setHeaders(headers: HeadersMap): void;
+
+    setHeader(key: string, value: string | string[]): void;
+
+    getCamelMessage(): typeof CamelMessage;
+}
--- a/components/api/api-modules-javascript/src/main/resources/META-INF/dirigible/modules/src/integrations/index.ts
+++ b/components/api/api-modules-javascript/src/main/resources/META-INF/dirigible/modules/src/integrations/index.ts
@@ -1,6 +1,7 @@
 const SpringBeanProvider = Java.type("org.eclipse.dirigible.components.spring.SpringBeanProvider");
 const Invoker = Java.type('org.eclipse.dirigible.components.engine.camel.invoke.Invoker');
 const invoker = SpringBeanProvider.getBean(Invoker.class);
+const String = Java.type('java.lang.String');
 
 export function invokeRoute(routeId, payload, headers) {
     return invoker.invokeRoute(routeId, payload, headers);
@@ -8,4 +9,55 @@ export function invokeRoute(routeId, payload, headers) {
 
 export function getInvokingRouteMessage() {
     return __context.camelMessage;
+}
+
+// Define an interface for the headers
+export interface Headers {
+    [key: string]: any;
+}
+
+// Wrapper class for Camel Message
+export class IntegrationMessage {
+    private message: any; // Replace 'any' with the actual type of Camel Message
+
+    constructor(message: any) {
+        this.message = message;
+    }
+
+    // Get the body of the message
+    getBody(): String {
+        return this.message.getBody(String.class);
+    }
+
+    // Set the body of the message
+    setBody(body: String): void {
+        this.message.setBody(body);
+    }
+
+    // Get headers of the message
+    getHeaders(): Headers {
+        const headers: Headers = {};
+        const headerNames = this.message.getHeaders().keySet().toArray();
+        for (const headerName of headerNames) {
+            headers[headerName] = this.message.getHeader(headerName);
+        }
+        return headers;
+    }
+
+    // Set headers of the message
+    setHeaders(headers: Headers): void {
+        for (const [key, value] of Object.entries(headers)) {
+            this.message.setHeader(key, value);
+        }
+    }
+
+    // Set header of the message
+    setHeader(key: String, value: any): void {
+        this.message.setHeader(key, value);
+    }
+
+    // Returns the original camel message
+    getCamelMessage(): any {
+        return this.message;
+    }
 }
--- a/components/api/api-modules-javascript/src/main/resources/META-INF/dirigible/modules/src/integrations/index.ts
+++ b/components/api/api-modules-javascript/src/main/resources/META-INF/dirigible/modules/src/integrations/index.ts
@@ -10,28 +10,3 @@ export function invokeRoute(routeId, payload, headers) {
 export function getInvokingRouteMessage() {
     return __context.camelMessage;
 }
-
-export interface HeadersMap {
-    [key: string]: string | string[];
-}
-
-export interface IntegrationMessage {
-
-    constructor(message: any);
-
-    getBody(): any;
-
-    getBodyAsString(): string;
-
-    setBody(body: any): void;
-
-    getHeaders(): HeadersMap;
-
-    getHeader(key: string): string | string[];
-
-    setHeaders(headers: HeadersMap): void;
-
-    setHeader(key: string, value: string | string[]): void;
-
-    getCamelMessage(): typeof CamelMessage;
-}
--- a/components/engine/engine-camel/src/main/java/org/eclipse/dirigible/components/engine/camel/invoke/IntegrationMessage.java
+++ b/components/engine/engine-camel/src/main/java/org/eclipse/dirigible/components/engine/camel/invoke/IntegrationMessage.java
@@ -1,0 +1,41 @@
+package org.eclipse.dirigible.components.engine.camel.invoke;
+
+import org.apache.camel.Message;
+
+import java.util.Map;
+
+public class IntegrationMessage {
+    private final Message message;
+
+    public IntegrationMessage(Message message) {
+        this.message = message;
+    }
+
+    public String getBodyAsString() {
+        return message.getBody(String.class);
+    }
+
+    public Object getBody() {
+        return message.getBody();
+    }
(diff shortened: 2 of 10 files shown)


components/engine/engine-camel/src/main/java/org/eclipse/dirigible/components/engine/camel/invoke/Invoker.java
```diff
@@ -11,37 +11,76 @@
...
-            runner.runMethod(module, "onMessage", messageBody);
+
+        var module = runner.run(Path.of(resourcePath));
+        var result = runner.runMethod(module, "onMessage", wrapCamelMessage(camelMessage));
+
+        camelMessage.getExchange()
+                    .setMessage(unwrapCamelMessage(result));
+    }
+
+    private Value wrapCamelMessage(Message camelMessage) {
```
vmutafov, line 50: As discussed offline, the implementation of this function would be changed

vmutafov (changes requested):

components/api/api-modules-javascript/src/main/resources/META-INF/dirigible/modules/src/integrations/index.ts
```diff
@@ -1,6 +1,7 @@
 const SpringBeanProvider = Java.type("org.eclipse.dirigible.components.spring.SpringBeanProvider");
 const Invoker = Java.type('org.eclipse.dirigible.components.engine.camel.invoke.Invoker');
 const invoker = SpringBeanProvider.getBean(Invoker.class);
+const CamelMessage = Java.type('org.apache.camel.Message');
```
vmutafov, line 4: Is this used?

components/engine/engine-camel/src/main/java/org/eclipse/dirigible/components/engine/camel/invoke/IntegrationMessage.java
```diff
@@ -0,0 +1,41 @@
+package org.eclipse.dirigible.components.engine.camel.invoke;
+
+import org.apache.camel.Message;
+
+import java.util.Map;
+
+public class IntegrationMessage {
```
vmutafov, line 7: Where's the TypeScript interface for this class?

components/engine/engine-camel/src/main/java/org/eclipse/dirigible/components/engine/camel/invoke/Invoker.java
```diff
@@ -11,37 +11,68 @@
...
+    private Message unwrapCamelMessage(Value value) {
+        if (value.isHostObject()) {
+            IntegrationMessage integrationMessage = value.asHostObject();
+            return integrationMessage.getCamelMessage();
+        } else {
+            throw new IllegalStateException("Unexpected @dirigible/integrations onMessage() return type");
         }
     }
 
+    private void resetCodeRunner() {
```
vmutafov, line 63: Have you checked whether `Invoker` is a request scoped bean?

components/engine/engine-camel/src/main/java/org/eclipse/dirigible/components/engine/camel/invoke/Invoker.java
```diff
@@ -11,37 +11,68 @@
...
+        camelMessage.getExchange()
+                    .setMessage(unwrapCamelMessage(result));
+    }
+
+    private IntegrationMessage wrapCamelMessage(Message camelMessage) {
+        return new IntegrationMessage(camelMessage);
+    }
+
+    private Message unwrapCamelMessage(Value value) {
+        if (value.isHostObject()) {
```
vmutafov, line 55: Invert the logic here to reduce nesting:
```Java
if (!value.isHostObject) {
 ...
} 
return ...
```

What happens if value is a host object but not an IntegrationMessage?
Also, specify in the exception what type exactly should be returned.

tests/src/test/java/org/eclipse/dirigible/integration/tests/services/integrations/CamelPlatformHttpTest.java
```diff
@@ -0,0 +1,52 @@
...
+import org.springframework.test.annotation.DirtiesContext;
+import org.springframework.test.web.servlet.MockMvc;
+import org.springframework.test.web.servlet.request.MockMvcRequestBuilders;
+
+import static org.junit.Assert.assertEquals;
+
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+@AutoConfigureMockMvc
+@DirtiesContext(classMode = DirtiesContext.ClassMode.AFTER_CLASS)
```
vmutafov, line 27: Do you need this?

tests/src/test/resources/META-INF/dirigible/camel_test/calledFromCamel.js
```diff
@@ -0,0 +1,7 @@
+exports.onMessage = (message) => {
```
vmutafov, line 1: Make this an `.mjs` file and use ESM exports

tests/src/test/resources/META-INF/dirigible/camel_test/calledFromCamel.js
```diff
@@ -0,0 +1,7 @@
...
+    let messageBody = message.getBodyAsString();
```
vmutafov, line 2: Use const

tests/src/test/resources/META-INF/dirigible/camel_test/calledFromCamel.js
```diff
@@ -0,0 +1,7 @@
...
+    let modifiedMessageBody = messageBody + " -> calledFromCamel.js handled this message";
```
vmutafov, line 3: Use const and string interpolation

tests/src/test/resources/META-INF/dirigible/camel_test/calledFromCamel.js
```diff
@@ -0,0 +1,7 @@
...
+    console.log('[CamelTest] CalledFromCamel.js called with message: ' + messageBody);
```
vmutafov, line 4: Do you need this?

tests/src/test/resources/META-INF/dirigible/modules-tests/src/services/integrations/dirigible-js-to-camel-route-test.js
```diff
@@ -0,0 +1,10 @@
+import { test, assertEquals } from "@dirigible/junit"
+import * as camel from "@dirigible/integrations"
+
+test('dirigible-js-to-camel-route-test', () => {
+    const message = "Initial Message";
+    const expected = message + " -> camel route inbound1 handled this message";
```
vmutafov, line 6: Use string interpolation

tests/src/test/resources/META-INF/dirigible/modules-tests/src/services/integrations/dirigible-js-to-camel-route-test.js
```diff
@@ -0,0 +1,10 @@
...
+	const actual = camel.invokeRoute('direct:inbound1', message, []);
+    console.log('[CamelTest] Camel route inbound1 responded with message: ' + actual);
```
vmutafov, line 8: Do you need this?

vmutafov (changes requested):

tests/src/test/resources/META-INF/dirigible/modules-tests/src/services/integrations/dirigible-js-to-camel-route-test.js
Fluctuationqt: removed the log

tests/src/test/resources/META-INF/dirigible/camel_test/calledFromCamel.js
Fluctuationqt: removed the log

tests/src/test/resources/META-INF/dirigible/camel_test/calledFromCamel.js
Fluctuationqt: changed to string interpolation and const

tests/src/test/resources/META-INF/dirigible/camel_test/calledFromCamel.js
Fluctuationqt: changed to const

tests/src/test/resources/META-INF/dirigible/camel_test/calledFromCamel.js
Fluctuationqt: changed to mjs with export function...

tests/src/test/java/org/eclipse/dirigible/integration/tests/services/integrations/CamelPlatformHttpTest.java
Fluctuationqt: Removed the dirties context

components/engine/engine-camel/src/main/java/org/eclipse/dirigible/components/engine/camel/invoke/Invoker.java
Fluctuationqt: inverted and added checks for null value, host object and instanceof IntegrationMessage, also a null check for the .getMessage() IntegrationMessage

components/engine/engine-camel/src/main/java/org/eclipse/dirigible/components/engine/camel/invoke/IntegrationMessage.java
Fluctuationqt: I removed it by mistake it's added back in the PR

components/api/api-modules-javascript/src/main/resources/META-INF/dirigible/modules/src/integrations/index.ts
Fluctuationqt: Yes in the TS interface

components/engine/engine-camel/src/main/java/org/eclipse/dirigible/components/engine/camel/invoke/Invoker.java
Fluctuationqt: From what i can see by default, the beans created with @Component are singleton scoped, meaning there will be only one instance per Spring context. There is @RequestScoped annotation that can be used in combination as described here https://stackoverflow.com/questions/74708265/component-with-request-scope-attributes if we need it

components/engine/engine-camel/src/main/java/org/eclipse/dirigible/components/engine/camel/invoke/Invoker.java
Fluctuationqt: Changed to Java wrapper

components/engine/engine-camel/src/main/java/org/eclipse/dirigible/components/engine/camel/invoke/Invoker.java
```diff
@@ -14,34 +14,76 @@
...
+
+        Object hostObject = value.asHostObject();
+
+        if (!(hostObject instanceof IntegrationMessage integrationMessage)) {
+            throw new IllegalArgumentException("Unexpected return type received from " + "@dirigible/integrations::onMessage(). "
+                    + "Expected return type: org.eclipse.dirigible.components.engine.camel.invoke.IntegrationMessage");
+        }
+
+        if (integrationMessage.getCamelMessage() == null) {
+            throw new IllegalStateException(
```
vmutafov, line 67: Wouldn't it work even if it's null?

components/engine/engine-camel/src/main/java/org/eclipse/dirigible/components/engine/camel/invoke/Invoker.java
```diff
@@ -14,34 +14,76 @@
...
+        camelMessage.getExchange()
+                    .setMessage(unwrapCamelMessage(result));
+    }
+
+    private IntegrationMessage wrapCamelMessage(Message camelMessage) {
+        return new IntegrationMessage(camelMessage);
+    }
+
+    private Message unwrapCamelMessage(Value value) {
+        if (value == null || !value.isHostObject()) {
```
vmutafov, line 54: Move those checks into a separate validation method

components/engine/engine-camel/src/main/java/org/eclipse/dirigible/components/engine/camel/invoke/Invoker.java
```diff
@@ -14,34 +14,76 @@
...
+            throw new IllegalStateException("Unexpected null value or return type received from " + "@dirigible/integrations::onMessage(). "
+                    + "Expected return type: org.eclipse.dirigible.components.engine.camel.invoke.IntegrationMessage");
+        }
+
+        Object hostObject = value.asHostObject();
+
+        if (!(hostObject instanceof IntegrationMessage integrationMessage)) {
+            throw new IllegalArgumentException("Unexpected return type received from " + "@dirigible/integrations::onMessage(). "
+                    + "Expected return type: org.eclipse.dirigible.components.engine.camel.invoke.IntegrationMessage");
```
vmutafov, line 63: Expected return type should probably be just IntegrationMessage as this is the TypeScript interface name. Users wouldn't use the Java type I guess

components/engine/engine-camel/src/main/java/org/eclipse/dirigible/components/engine/camel/invoke/Invoker.java
vmutafov: Request scope is not the only alternative here, it could also be prototype scoped. If you're using the `Invoker` from the Camel Bean component (and not from the regular Class component) you could try and see if Camel requests a new bean every time or it caches it somehow. If using prototype scope results in new `Invoker` instances for every invocation from Camel, then we can tie the lifecycle of the GraalJS context to the Spring Bean lifecycle - https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/annotation/Bean.html#destroyMethod()

vmutafov (changes requested):

components/engine/engine-camel/src/main/java/org/eclipse/dirigible/components/engine/camel/invoke/Invoker.java
Fluctuationqt: Changed to IntegrationMessage

components/engine/engine-camel/src/main/java/org/eclipse/dirigible/components/engine/camel/invoke/Invoker.java
Fluctuationqt: it would, the only time we need to check for null should be on the result of the js call as in that case we should not modify the message in the exchange. I added that check where we get the result so that we don't call any message change logic

components/engine/engine-camel/src/main/java/org/eclipse/dirigible/components/engine/camel/invoke/Invoker.java
Fluctuationqt: done

components/engine/engine-camel/src/main/java/org/eclipse/dirigible/components/engine/camel/invoke/Invoker.java
Fluctuationqt: I tested it out by having the Invoker class scoped with @Scope("prototype"). In the debuger i can see that in the **Invoker::invoke()** method **this** is the same object for every http request when in the designer it's singleton and when it's prototype in the designer it's always a different object

vmutafov: And have you tried making the `Invoker` implement `AutoCloseable` and see if the close method is invoked as expected?

vmutafov (approved):

ThuF (approved):
+1 Seems good to me.

PR #3174, opened 2024-01-03
fix: use custom repository via DIRIGIBLE_REPOSITORY_LOCAL_ROOT_FOLDER

iliyan-velichkov:


iliyan-velichkov:
Preventive measure for this issue: https://github.com/eclipse/dirigible/issues/3175

iliyan-velichkov:
@ThuF @delchev @vmutafov the pr is for review

components/core/core-repository/src/main/java/org/eclipse/dirigible/components/repository/RepositoryConfig.java
```diff
@@ -10,20 +10,26 @@
...
-@Configuration
+@org.springframework.context.annotation.Configuration
 public class RepositoryConfig {
 
+
     @Bean
     public IRepository repository() {
-        LocalRepository localRepository = new LocalRepository("target");
+        String repoFolderPath = Configuration.get(Configuration.DIRIGIBLE_REPOSITORY_LOCAL_ROOT_FOLDER, "target");
+        boolean absolutePath = repoFolderPath.startsWith(File.separator);
```
delchev, line 27: let use the explicit DIRIGIBLE_REPOSITORY_LOCAL_ROOT_FOLDER_IS_ABSOLUTE here as to work on any filesystem

delchev (changes requested):

components/core/core-repository/src/main/java/org/eclipse/dirigible/components/repository/RepositoryConfig.java
iliyan-velichkov: It uses the files separator to check whether it is absolute.  This way it should be working on every os or I'm missing something?

delchev: on Windows you could have something like `C:\MyFolder` which will not be taken as an absolute

iliyan-velichkov: Somehow I forgot about the Windows :)
I will fix it on Monday.

delchev (approved):

PR #3198, opened 2024-01-09
Do not fail on missing config for property with multi placeholders

iliyan-velichkov:
- Do not fail on missing config for property with multi placeholders
- Add more details on some exceptions


iliyan-velichkov:
@delchev the PR is for review

delchev (approved):

PR #3264, opened 2024-01-21
Bar Chart

delchev:


ThuF:
Implemented in: https://github.com/eclipse/dirigible/pull/3548

PR #3351, opened 2024-01-31
Fix logging

iliyan-velichkov:


iliyan-velichkov:
@delchev @ThuF @vmutafov the PR is for review

Issue #3358, opened 2024-02-01
[EDM] Cell value is not applied on enter


StanZGenchev:
**Describe the issue**
When you open an edm file and rename a key inside an entity, when you hit enter, a new line is created. The expected behavior is that the new name to be applied. New lines are not even permitted for key names.

**To Reproduce**
1. Open an EDM file
2. Start renaming a key inside an entity
3. Hit enter
4. See issue

**Expected behavior**
The new key name should be applied on enter.

**Screenshots**
![image](https://github.com/eclipse/dirigible/assets/82379274/fb586f7c-7d78-4d1b-a869-b12fcd3bdae4)

**Desktop (please complete the following information):**
 - OS: Fedora Linux, macOS, Windows
 - Browser: All
 - Version Dirigible 9.2.1

PR #3360, opened 2024-02-01
[EDM] #3358 - Cell value is not applied on enter

StanZGenchev:
### What does this PR do?

Updated the Entity editor config, so that when a user is renaming the cell value and hits enter, the value will be applied instead of creating a new line.

#3358


ThuF (approved):
+1 Seems good to me.

PR #3362, opened 2024-02-02
Fix column names escaping in where statements

iliyan-velichkov:


iliyan-velichkov:
@delchev @ThuF the PR is for review

ThuF (approved):
+1 Seems good to me.

Issue #3370, opened 2024-02-05
[Database] Export endpoints to export to default workspace


nedelcho-delchev-tues:
Some endpoints for exporting different database data are currently exporting to `PUBLIC` workspace in `PUBLIC` project.
We want to export in the default `workspace` if present if not create it.

Issue #3369, opened 2024-02-05
[Template] Make TablePrefix optional when generating application


nedelcho-delchev-tues:
Related to:
- https://github.com/eclipse/dirigible/issues/3353

PR #3371, opened 2024-02-05
[#3353] Generate model from schema

### LINKED ISSUE #3353: [EDM] Generate Model from Schema
From the Database Explorer would be good to add an action under Schema node for generation a model from the selected schema. It will be useful when afterwards the developer can generate data access layer or even up to the API layer (or even simple UI)

nedelcho-delchev-tues:
Right now the generate project needs to be edited a little because of the mandatory table prefix when generating an application. This will be fix with #3369 

Quick fix:
- in gen/dao/your_tables.js remove the prefix in the table string to be the same as the table in the datasource

fixes #3353, #3370


ThuF (approved):
+1 Seems good to me.

Issue #3378, opened 2024-02-05
[Schema] Add Suport for Data Source Different from the DefaultDB


ThuF:
With the introduction of the following, now it's possible to interact with tables from different data sources _(e.g. `SystemDB` and/or custom defined ones)_:
- https://github.com/eclipse/dirigible/pull/3377
- https://github.com/eclipse/dirigible/issues/3369
- https://github.com/eclipse/dirigible/issues/3368

This is working fine for entities, already existing in the specified data source _(e.g. `SystemDB` and/or custom defined ones)_. However if a new entity (database table) has to be crated into the specified data source, it's being created in the `DefaultDB`.

We should add support for optional `dataSource` property in the `*.schema` files and support for it during synchronization.

ThuF:
Partial implementation _(data source for the whole module)_ done with:
- https://github.com/eclipse/dirigible/pull/3389

PR #3389, opened 2024-02-07
[Schema] Support for different datasource for whole model

nedelcho-delchev-tues:
Part of:
- #3378 

The `.schema` file when generating application from `.model` now has `dataSource` _(`DefaultDB` is still default if no datasource is given as an input)_ and is taken in mind when doing CRUD operations from the EDM, eg - new entity will be created in the chosen dataSource instead of `DefaultDB`.

Next part will be to implement support for each entity to be able to have different datasource.


ThuF (approved):
+1 Seems good to me.

delchev (approved):

PR #3405, opened 2024-02-08
[Camel] Cannot execute TypeScript files

ThuF:
Contribution PR:
- https://github.com/eclipse/dirigible/pull/3395


modules/engines/engine-graalium/execution-core/src/main/java/org/eclipse/dirigible/graalium/core/DirigibleJavascriptCodeRunner.java
```diff
@@ -178,8 +182,7 @@ public Module run(Path codeFilePath) {
 
     public Value runMethod(Module codeModule, String methodName, Object... args) {
         return switch (codeModule.moduleType()) {
-            case CJS -> runCjsMethod(codeModule.module(), methodName, args);
-            case ESM -> runEsmMethod(codeModule.module(), methodName, args);
+            case CJS, ESM -> runEsmMethod(codeModule.module(), methodName, args);
```
vmutafov, line 185: Why is this necessary?

modules/engines/engine-graalium/execution-core/src/main/java/org/eclipse/dirigible/graalium/core/DirigibleJavascriptCodeRunner.java
```diff
@@ -241,10 +236,7 @@ public GraalJSInterceptor getGraalJSInterceptor() {
     }
 
     private static String transformTypeScriptHandlerPathIfNecessary(String handlerPath) {
-        if (handlerPath.endsWith(TS_EXT)) {
-            return handlerPath.substring(0, handlerPath.length() - TS_EXT.length()) + MJS_EXT;
-        }
-        return handlerPath;
+        return handlerPath.endsWith(TS_EXT) ? handlerPath.replaceAll(Pattern.quote(TS_EXT), JS_EXT) : handlerPath;
```
vmutafov, line 239: This is a breaking change in some cases. Why are we changing the extension to “.js”?

modules/engines/engine-graalium/execution-core/src/main/java/org/eclipse/dirigible/graalium/core/DirigibleJavascriptCodeRunner.java
vmutafov: This is incorrect. If there is a problem somewhere, it’s definitely not here. CommonJS and ESM are two completely different module systems and there is no standard way for interoperability between them. We shouldn’t treat CommonJS modules as ESMs.

vmutafov (changes requested):

modules/engines/engine-graalium/execution-core/src/main/java/org/eclipse/dirigible/graalium/core/DirigibleJavascriptCodeRunner.java
ThuF: Reverted to the original content and execution flow

modules/engines/engine-graalium/execution-core/src/main/java/org/eclipse/dirigible/graalium/core/DirigibleJavascriptCodeRunner.java
ThuF: With the latest commit the file extension is still changed from `*.ts` to `*.js`, however in the case of `*.ts` the **ModuleType** is assigned to `ModuleType.ESM`

vmutafov (approved):

PR #3421, opened 2024-02-09
[EDM] Support Filtering for Entities

ThuF:
**Note:** With this PR filtering is introduced for all layouts except the "master view" of the "master-detail" layout.

Filtering demo:

https://github.com/eclipse/dirigible/assets/4092083/b00e4680-7f97-4a06-9372-4bdf86b82ea2


iliyan-velichkov:
@ThuF do we support filtering of the master entities? (filter next to Create, Edit, Delete)
image

Issue #3147, opened 2023-12-23
[EDM] Introduce ReadOnly Entity


delchev:
Introduce `ReadOnly` Entity
* Will not be included in the schema generation file
* Will not have generated CUD but only READ and LIST methods in DAO/API

The idea behind is that the original Entity is owned by another model where full lifecycle happens. Here the ReadOnly Entity plays a kind of projection e.g. with very few fields included.

delchev:
Projection entity will serve this need: https://github.com/eclipse/dirigible/commit/2d998a715fd20024f722a0943c2697db01ff897e

delchev:
Addressed by projection entities, per the comment above and #3422 (merged 2024-02-13). A `PROJECTION` entity is exactly the read-only shape described here — excluded from schema generation and generated without CUD — and the support is on master across the EDM, `template-application-schema` and the intent generator. Closing as resolved; please reopen if the ReadOnly semantics intended here differ from what projection entities provide.

PR #3422, opened 2024-02-09
[EDM] Support for projection entities

nedelcho-delchev-tues:
- Added checkbox `Disable generation` in the EDM for entities to stop generation of files for the selected entitiy
- Changed the templete for `controller.js` to now correctly get the `Service.ts` component for projection entities from their own projects

### DIFF
--- a/components/template/template-application-ui-angular/src/main/resources/META-INF/dirigible/template-application-ui-angular/ui/perspective/manage/controller.js.template
+++ b/components/template/template-application-ui-angular/src/main/resources/META-INF/dirigible/template-application-ui-angular/ui/perspective/manage/controller.js.template
@@ -172,19 +172,47 @@ angular.module('page', ["ideUI", "ideView", "entityApi"])
 		$scope.options${property.name} = [];
 #end
 #end
-#foreach ($property in $properties)
-#if($property.widgetType == "DROPDOWN")
 
-		${dollar}http.get("/services/ts/${projectName}/gen/api/${property.relationshipEntityPerspectiveName}/${property.relationshipEntityName}Service.ts").then(function (response) {
-			${dollar}scope.options${property.name} = response.data.map(e => {
-				return {
-					value: e.${property.widgetDropDownKey},
-					text: e.${property.widgetDropDownValue}
-				}
-			});
-		});
+#foreach ($property in $properties)
+#if ($property.widgetType == "DROPDOWN")
+#if ($referencedProjections.length != 0)
+#set ($foundReferenceProjection = false)
+#foreach ($referencedProjection in $referencedProjections)
+                #if ($referencedProjection.name == $property.relationshipEntityName && !$foundReferenceProjection)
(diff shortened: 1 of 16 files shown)


delchev (approved):

components/template/template-application-ui-angular/src/main/resources/META-INF/dirigible/template-application-ui-angular/ui/perspective/manage/controller.js.template
```diff
@@ -172,19 +172,47 @@ angular.module('page', ["ideUI", "ideView", "entityApi"])
...
-					text: e.${property.widgetDropDownValue}
-				}
-			});
-		});
+#foreach ($property in $properties)
+#if ($property.widgetType == "DROPDOWN")
+#if ($referencedProjections.length != 0)
+#set ($foundReferenceProjection = false)
+#foreach ($referencedProjection in $referencedProjections)
+                #if ($referencedProjection.name == $property.relationshipEntityName && !$foundReferenceProjection)
```
ThuF, line 181: The spaces in the beginning could be removed:
```
                #if ($referencedProjection.name == $property.relationshipEntityName && !$foundReferenceProjection)
```
->
```
#if ($referencedProjection.name == $property.relationshipEntityName && !$foundReferenceProjection)
```

components/template/template-application-ui-angular/src/main/resources/META-INF/dirigible/template-application-ui-angular/ui/perspective/manage/controller.js.template
```diff
@@ -172,19 +172,47 @@ angular.module('page', ["ideUI", "ideView", "entityApi"])
...
+              ${dollar}scope.options${property.name} = response.data.map(e => {
+                  return {
+                      value: e.${property.widgetDropDownKey},
+                      text: e.${property.widgetDropDownValue}
+                  }
+              });
+          });
 #end
 #end
+#end
```
ThuF, line 214: I think it would be a better for the whole logic to happen inside the `parameterUtils.js`, as here in the template it's too hard to understand it. In the end only the url is different:
```
"/services/ts/${projectName}/gen/api/${property.relationshipEntityPerspectiveName}/${property.relationshipEntityName}Service.ts"
```
vs
```
"/services/ts/${referencedProjection.project}/gen/api/${property.relationshipEntityPerspectiveName}/${property.relationshipEntityName}Service.ts"
```

ThuF (commented):
The `dataSource` properties were removed with https://github.com/eclipse/dirigible/commit/2976a3a135f2550ae3854dbaa209d5c9b7a60ecb, maybe the PR needs to be rebased?

I think it would be better to refactor the logic for determining the dropdown url and maybe move it into the `parameterUtils.js` in order to simplify the content and the logic that happens inside the templates.

PS: I guess the same change should be adopted also for the `list` and `master-list` templates?

ThuF:
@nedelcho-delchev-tues, @delchev does this PR closes this issue?
- https://github.com/eclipse/dirigible/issues/3147

ThuF (approved):
+1 Seems good to me! @nedelcho-delchev-tues, should we support the same functionality also for the "list" and "master-list" templates?

PR #3462, opened 2024-02-18
Cannot create boolean column which is not null

delchev:
Cannot create boolean column which is not null


delchev:
Screenshot 2024-02-18 at 10 33 51
Screenshot 2024-02-18 at 10 34 03
Screenshot 2024-02-18 at 10 34 16

PR #3593, opened 2024-03-15
Introduce Cache Api

MinaDoncheva:


components/api/api-modules-javascript/src/main/resources/META-INF/dirigible/modules/src/cache/caches.ts
```diff
@@ -0,0 +1,27 @@
...
+ */
+const Cache = Java.type("org.eclipse.dirigible.components.api.cache.CacheFacade");
+
+export class Caches {
+
+    public static contains(key: string): boolean {
+        return (Cache.contains(key));
+    }
+
+    public static get(key: string): string | undefined {
```
ThuF, line 12: I think the return type should be `any | undefined` instead of `string | undefined`, as the `get()` method in the `CacheFacade` returns `Object`.

components/api/api-modules-javascript/src/main/resources/META-INF/dirigible/modules/src/cache/caches.ts
```diff
@@ -0,0 +1,27 @@
+/**
+ * API Cache
+ */
+const Cache = Java.type("org.eclipse.dirigible.components.api.cache.CacheFacade");
```
ThuF, line 4: Either all methods in the `CacheFacade` should be `static` (in order to be used in the JS/TS like this `Cache.contains()`), or this should be changed to:

```ts
const CacheFacade = Java.type("org.eclipse.dirigible.components.api.cache.CacheFacade");
const cache = new CacheFacade();

export class Cache {

    public static contains(key: string): boolean {
        return cache.contains(key);
    }
...
```

components/api/api-modules-javascript/src/main/resources/META-INF/dirigible/modules/src/cache/caches.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export class Caches {
```
ThuF, line 6: I guess `Cache` is better name than `Caches`?

components/api/api-cache/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+		<dependency>
+			<groupId>org.eclipse.dirigible</groupId>
+			<artifactId>dirigible-components-api-test</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>com.github.ben-manes.caffeine</groupId>
+			<artifactId>caffeine</artifactId>
+		</dependency>
+
+	</dependencies>
```
ThuF, line 43: Please double-check whether all dependencies are needed

components/api/api-cache/pom.xml
```diff
@@ -0,0 +1,50 @@
...
+
+	<parent>
+		<groupId>org.eclipse.dirigible</groupId>
+		<artifactId>dirigible-components-parent</artifactId>
+		<version>11.0.0-SNAPSHOT</version>
+		<relativePath>../../pom.xml</relativePath>
+	</parent>
+
+	<name>Components - API - Cache</name>
+	<artifactId>dirigible-components-api-cache</artifactId>
```
ThuF, line 12: I think the `dirigible-components-api-cache` module, should be added as dependency in one of those, in order to be included in the executable JAR:
- https://github.com/eclipse/dirigible/blob/master/components/group/group-api/pom.xml
- https://github.com/eclipse/dirigible/blob/master/components/group/group-api-core/pom.xml
- https://github.com/eclipse/dirigible/blob/master/components/group/group-api-core/pom.xml

ThuF (changes requested):

ThuF (approved):
+1 Seems good to me.

delchev (approved):

PR #3609, opened 2024-03-19
Multitenancy phase 3 - listeners

iliyan-velichkov:
Implementation of https://github.com/eclipse/dirigible/discussions/3423 [phase 3]
Add multitenancy support for listeners (`*.listener`) and messaging API.
- send messaging to dedicated topic/queue depending on the current tenant
   - for the default tenant topic/queue name is used as it is
   - for non defualt tenant - a prefix with the current tenant id is added to the topic/queue name
   - an example for topic with name `myTopic`
      - for the default tenant the name will be `myTopic`
      - for a tenant with id `148e34d8-47ff-4e53-9f74-f0100202fdbf` the name will be `148e34d8-47ff-4e53-9f74-f0100202fdbf###myTopic`
- execute listener handlers in the corresponding tenant context
- register a dedicated  message handlers for all provisioned tenants

Example logs for executing [this](https://github.com/iliyan-velichkov/dirigible-test-project/blob/master/dirigible-test-project/listeners/entity/book-entity-events.listener) test listener on creating new Book from the UI.
image

### DIFF
--- a/components/engine/engine-listeners/src/main/java/org/eclipse/dirigible/components/listeners/service/Listener.java
+++ b/components/engine/engine-listeners/src/main/java/org/eclipse/dirigible/components/listeners/service/Listener.java
@@ -1,0 +1,70 @@
+package org.eclipse.dirigible.components.listeners.service;
+
+import org.eclipse.dirigible.components.base.spring.BeanProvider;
+import org.eclipse.dirigible.components.listeners.domain.ListenerKind;
+
+import java.util.Objects;
+
+class Listener {
+    private final ListenerType type;
+    private final String destination;
+    private final String handlerPath;
+
+    Listener(ListenerType type, String destination, String handlerPath) {
+        this.type = type;
+        this.destination = destination;
+        this.handlerPath = handlerPath;
+    }
+
+    ListenerType getType() {
+        return type;
+    }
+
+    String getDestination() {
+        return destination;
+    }
+
+    String getHandlerPath() {
+        return handlerPath;
+    }
+
+    @Override
+    public String toString() {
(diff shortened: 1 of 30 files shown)


iliyan-velichkov:
@delchev @vmutafov @ThuF the PR is ready for review

components/engine/engine-listeners/src/main/java/org/eclipse/dirigible/components/listeners/service/Listener.java
```diff
@@ -0,0 +1,48 @@
+package org.eclipse.dirigible.components.listeners.service;
+
+import java.util.Objects;
+
+class Listener {
```
delchev, line 5: isn't it ListenerService?

components/engine/engine-listeners/src/main/java/org/eclipse/dirigible/components/listeners/service/Listener.java
```diff
@@ -0,0 +1,48 @@
...
+    private final ListenerType type;
+    private final String destination;
+    private final String handlerPath;
```
delchev, line 8: can we keep just `handler` here as all the rest?

delchev (changes requested):
naming conventions only

components/engine/engine-listeners/src/main/java/org/eclipse/dirigible/components/listeners/service/Listener.java
iliyan-velichkov: renamed to ListenerDescriptor

delchev (approved):

Issue #3604, opened 2024-03-18
[BPM][Inbox] Implement list process task REST endpoint(s)


art-entrery:
In order to show all the relevant tasks for the currently logged-in user we need to implement REST endpoint as follows:

1. Get the currently active user from the Spring session
2. List all the roles of the currently active user
3. Based on the roles return all the relevant tasks where the user is one of the following: assignee, candidate user, part of candidate user group

NOTE: The key insight here is that we need a match between the user or group assigned during the process modeling (design time) and the roles assigned to the currently logged in user

ThuF:
@art-entrery Should we consider this issue done?

entrery:
Done with #3621

PR #3621, opened 2024-03-20
[BPM][Inbox] Feature#3604: Implement get tasks API using security context

entrery:
This PR is implementation for FR #3604 
The idea is to take the roles or username of the currently logged-in user and based on that to retrieve all the user tasks assigned (or candidate) to him


entrery:
@delchev

delchev (approved):

PR #3655, opened 2024-03-27
[BPM][Inbox] Claim user task functionality

entrery:


entrery:
@delchev

entrery:
@delchev Is it possible to retrigger individual failed tests as I am not sure whether the problem is caused by my implementation

delchev (approved):

Issue #3551, opened 2024-03-07
Usability improvement on entering entity data using the keyboard only


iliyan-velichkov:
When I'm entering data for a new entity, I should be able to switch between fields using single tab not using two tabs.
For example:
If my focus is on Title
![image](https://github.com/eclipse/dirigible/assets/5058839/99271877-4d0d-4717-abbc-c9d5d2cb66cf)
I want to be able to click tab and go to Publisher.
Currently, two tabs are required.
The same applies when I want to go back (shift + tab).

PR #3698, opened 2024-04-02
[UI] #3551 - Usability improvement on entering entity data using the keyboard only

StanZGenchev:
### What does this PR do?

Improves widget navigation while using a keyboard input.

#3551


ThuF (approved):
+1 Seems good to me.

PR #3741, opened 2024-04-08
skip misfired jobs executions

iliyan-velichkov:
- reschedule system jobs on startup to apply new changes if any
- skip system misfired job executions
- add quartz startup delay to prevent jobs execution before spring boot application startup 
- optimize SynchronizationJob
  - remove explicit calls for garbage collection - it causes Dirigible slow down when the job is executed frequently (for example every 100 millis)
  - remove ExecutorService since it not needed
  - optimize logging


iliyan-velichkov:
"Pull Request / build (windows) (pull_request)" fails because tests are broken on windows in the master ---> https://github.com/eclipse/dirigible/pull/3751

Issue #3782, opened 2024-04-13
[Templates] Support of Widget Size in the Form Designer


delchev:
Support of widget size in the Form Designer like in the EDM Modeller for UI

Issue #3761, opened 2024-04-09
[Form] Paragraph element is deleted when pressing the Delete button inside the Text property


ThuF:
Pressing the `Delete` button on the keyboard while editing the `Text` property of the **Paragraph** element results into the **Paragraph** element being deleted from the model. The expected behaviour would be to delete the content inside the `Text` property and not the element itself.

The workaround is to use the Backspace button instead.

The expected behaviour would be when the `Text` property is in focus, then pressing the Delete button shouldn't delete the element.

**Edit:** It seems that it's the same behaviour for all elements.

ThuF:
@StanZGenchev I've tested the latest changes, everything seems to be working as expected, except the **Paragraph** widget. To test it you can:
- Create a new project from the `Form Definition (UI)` template _(from the `Welcome` page)_.
- Open the `order.form` file.
- Scroll down to the **Paragraph** element and click on it.
- If I press the `Delete` button on the keyboard, while editing the `Text` property, then the whole **Paragraph** element is deleted from the form

<img width="1471" alt="Screenshot 2024-04-30 at 16 03 00" src="https://github.com/eclipse/dirigible/assets/4092083/ec7c9ad5-593f-44a6-9e1a-2c7a6ec99df1">

PR #3879, opened 2024-05-01
[Templates] #3782 - Support of Widget Size in the Form Designer

StanZGenchev:
### What does this PR do?

Adds support for compact mode in the Form Editor and the template used for form generation.
Fixes paragraph getting deleted when a user presses the delete button while editing its text in the Form Editor.

#3782 
#3761


himchev (approved):
Looks good

PR #3986, opened 2024-06-03
fix tables and views check for existence

iliyan-velichkov:
Currently:
-  the existence check, doesn't include the artefact schema
- drop view doesn't support specifying the schema


iliyan-velichkov:
@delchev @ThuF the PR is for review

delchev (approved):

PR #4034, opened 2024-06-12
Add config props for leaked connections check interval and max in use time

iliyan-velichkov:
- Add config props for leaked connections check interval and max in use time
- Eagerly init the checks

New configs:
- `LEAKED_CONNECTIONS_MAX_IN_USE_SECONDS` - 180 seconds (3 mins) by default
- `DIRIGIBLE_LEAKED_CONNECTIONS_CHECK_INTERVAL_SECONDS` - 30 seconds


iliyan-velichkov:
@delchev the PR is for review

components/data/data-sources/src/main/java/com/zaxxer/hikari/pool/InUseConnectionEntry.java
```diff
@@ -0,0 +1,43 @@
+package com.zaxxer.hikari.pool;
```
delchev, line 1: Is there a special reason to use foreign package name? It will be a problem on publishing to Maven Central

delchev (changes requested):

components/data/data-sources/src/main/java/com/zaxxer/hikari/pool/InUseConnectionEntry.java
iliyan-velichkov: yes, package visibility of used  hikari classes

iliyan-velichkov: more specifically - com.zaxxer.hikari.pool.ProxyConnection#getPoolEntry

delchev: ok, thanks

PR #4043, opened 2024-06-13
ABAP - support inline statements

iliyan-velichkov:
related to: https://github.com/codbex/codbex-kronos/issues/598


iliyan-velichkov:
@delchev the PR is ready for review

PR #4111, opened 2024-07-03
BPM enhancements 

iliyan-velichkov:
- Add BPM Starter project template
   image
   image
   image
   image
   image
   image
   image

- Add TS API for getting all process variables
```
import { process } from "sdk/bpm"

const execution = process.getExecutionContext();
const executionId = execution.getId();

const processVariables = process.getVariables(executionId);
```
- Allow executing injectable spring tasks in Flowable processes. The bean name must be specified in the `Delegate expression` field of the system task. Example:
image

- Make `org.eclipse.dirigible.components.engine.bpm.flowable.delegate.DirigibleCallDelegate` a spring bean with name `JSTask`. 
   - This allows decoupling from the class reference, allowing to move it around the codebase without breaking the existing processes.
   - It is easier for the clients to define a JS task by putting the bean name `JSTask` in `Delegate expression` and providing the JS file path in `handler` field

   image
   image
   image

- The change is backward compatible


delchev (approved):
Awesome!

PR #4131, opened 2024-07-09
inbox APIs accessible by all authenticated users

iliyan-velichkov:
- Make inbox APIs accessible by every authenticated user not just the Developers. This allows every user to check his/her tasks.
   - Affected APIs
      - GET `/services/bpm/bpm-processes/tasks`
      - POST `/services/bpm/bpm-processes/tasks/*`
- Fix `UserFacade.getUserRoles` - remove `ROLE_` prefix when extracting the roles from the authorities since the prefix is spring specific
- do not allow processing tasks without appropriate permissions


delchev (approved):
Awesome!

PR #4133, opened 2024-07-10
[custom security] Ant path patterns in constraints

iliyan-velichkov:
Support adding ant path patterns in constraints.

Instead of the previous version which allows only exact paths, I need to protect APIs for paths like `/services/ts/leave-request/api/ProcessService.ts/requests/*/decline`

 Constraint UI was changed from
image

to

image

The request validation was adapted to use AntPathMatcher.


iliyan-velichkov:
@delchev the PR is for review

delchev (approved):
Awesome!

Issue #4134, opened 2024-07-10
Forms doesn't load properly in Chrome incognito mode and Safari


iliyan-velichkov:
Chrome Incognito mode
![image](https://github.com/eclipse/dirigible/assets/5058839/53851e3f-6d18-49f4-95f1-2a50be547935)
Video:
https://github.com/eclipse/dirigible/assets/5058839/2d9a46d6-9a10-42f3-9fe5-01ac7afd599e

vs

![image](https://github.com/eclipse/dirigible/assets/5058839/787b1599-6174-482f-bb86-66c5a26d67cb)

Safari:
![image](https://github.com/eclipse/dirigible/assets/5058839/839b6a91-2e10-4562-9419-97fea368baad)

iliyan-velichkov:
Findings:
- if I firstly open the IDE `http://localhost/services/web/ide/`, the page loads correctly
![image](https://github.com/eclipse/dirigible/assets/5058839/0f78deb4-f6c0-42bc-b49b-a57000c49157)

iliyan-velichkov:
https://github.com/eclipse/dirigible/assets/5058839/2d9a46d6-9a10-42f3-9fe5-01ac7afd599e
vs
https://github.com/eclipse/dirigible/assets/5058839/a4cc65c2-e086-4a72-8927-137d634a23d2

StanZGenchev:
This is not a problem with forms or with incognito mode at all. It affects all views.
The "problem" is that in order to have proper theming, the `theme` directive gets the theme config from local storage.
If there is no default config, it will get an empty object.
The default theme config is created in `core-modules` by calling the `theming` provider upon initial Dirigible load, not by the views.
And this is where the true "problem" is. Dirigible views are not meant to be viewed separately outside of Dirigible, so if you run them in a separate tab in a "clean" environment, the defaults will never be set.
If we want a view (in this case a form view) that uses Dirigibles UI, but is not a Dirigible view itself, to work outside of Dirigible, then we can inject the `theming` provider in the `Theme` factory inside `components/resources/resources-core/src/main/resources/META-INF/dirigible/resources-core/ui/theming.js`.

PR #4137, opened 2024-07-10
[UI] #4134 - Set default theme config

StanZGenchev:
### What does this PR do?

Makes a small change, where it injects the `theming` provider in the `Theme` factory, so that the default theme config can be set even if Dirigible itself has never been loaded before.

#4134


delchev (approved):

iliyan-velichkov (commented):
I verified the change and it works.

PR #4138, opened 2024-07-11
Fix mail api

iliyan-velichkov:
Fix maill JS API. Not it works and you can send emails.


iliyan-velichkov:
@delchev the PR is for review

delchev (approved):

PR #4157, opened 2024-07-19
Add workflow job for integration tests with PostgreSQL + fix issues related to PostgreSQL usage for DefaultDB

iliyan-velichkov:
Motivation: to ensure that the functionality works with PostgreSQL for DefaultDB

- configure workflows to work with ubuntu
- fix the code in the places which were not compatible with PostgreSQL for DefaultDB
- resolve tests failures caused by different tests execution order on ubuntu

PR #4169, opened 2024-07-24
Integration tests on ubuntu

iliyan-velichkov:


iliyan-velichkov:
The PR is failing due to issue in the tests execution caused by different ordering on ubuntu.
Will be implemented with https://github.com/eclipse/dirigible/pull/4157

The PR is used for testing as well.

Issue #3878, opened 2024-05-01
Docker image does not come up when using a mounted volume


g051051:
**Describe the bug**
Running the dirigiblelabe/dirigible:10.4.0 image fails to fully start when using a mounted volume for the target directory.

**To Reproduce**
1. Run the docker image as described in the [https://www.dirigible.io/help/setup/docker/](https://www.dirigible.io/help/setup/docker/) "with Mounted Volume" section.  Note that the instructions are incorrect and the -v option should be `-v <your-local-directory>:/target -p 8080:8080 -p 8081:8081`.
2. Open localhost:8080 in a browser
3. Log in as admin/admin
4. You should see a "busy" page indefinitely and an error in the log

**Expected behavior**
The system runs normally with the mounted volume.

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Chrome
 - Version: 124.0.6367.119 (Official Build) (64-bit)

**Additional context**
```
2024-05-01 08:33:05 2024-05-01 12:33:05.768 [ERROR] [http-nio-8080-exec-4] [background] o.a.c.c.C.[.[.[.[dispatcherServlet] - Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception [Request processing failed: java.lang.RuntimeException: java.nio.file.FileAlreadyExistsException: /target/dirigible/repository/root/registry/public/modules/dist/esm/http/rs] with root cause
2024-05-01 08:33:05 org.graalvm.polyglot.PolyglotException: java.nio.file.FileAlreadyExistsException: /target/dirigible/repository/root/registry/public/modules/dist/esm/http/rs
```

g051051:
It's not the loading screen issue.  It might be the Git line ending thing.  I'll test it again now that I've changed that setting and see if it makes a difference.

ThuF:
Not yet, we'll investigate it under windows too. In the meantime could you try to run it via the Windows WSL?

g051051:
I don't know how to run docker stuff under WSL, sorry.

g051051:
Pull 10.4.0 from docker hub:

![image](https://github.com/eclipse/dirigible/assets/2124974/d2e2a69a-14f0-44e9-9f3e-9b46687cdd91)

Run it:

![image](https://github.com/eclipse/dirigible/assets/2124974/8b798746-8742-4359-ac85-116712381b0c)

Log file (this is after the system had been sitting for 10 minutes):
[...]

PR #4229, opened 2024-08-16
Do not fail on missing resource

iliyan-velichkov:


iliyan-velichkov:
this change will not help to resolve the windows docker issue with mounted path.
Check [this comment](https://github.com/eclipse/dirigible/issues/3878#issuecomment-2293275645) for more details.

PR #4315, opened 2024-09-26
Authenticate using Snowflake authentication mechanism 

iliyan-velichkov:
When the Dirigible is deployed on Snowflake as container service, users should be able to authenticated in the application with Snowflake authentication mechanism only. Currently, double authentication is used.
 
The PR adds a new spring profile for Snowflake called `snowflake`.
The profile relies on the propagated logged in user for the authentication as described in [this](https://quickstarts.snowflake.com/guide/build_a_data_app_and_run_it_on_Snowpark_container_services/index.html#3) tutorial. Additional details [here](https://docs.snowflake.com/en/developer-guide/snowpark-container-services/working-with-services#user-specific-headers-in-ingress-requests).

Example env variable: `SPRING_PROFILES_ACTIVE=snowflake`

An additional **optional** env variable was introduced called `DIRIGIBLE_SNOWFLAKE_ADMIN_USERNAME`.
Using this configuration, an user can be automatically initialized with all available Dirigible roles. Later on, you can use this user to assign roles to other users.

When a user loggin to the application for the first time, a new user registration is automatically created in the system DB if missing.

I didn't find a proper way to implement logout. I have created a Snowflake ticket for that.  

I verified [the Snowflake UDF scenario](https://www.codbex.com/technology/2024/09/18/snowflake-udf-leave-days/) works with the current implementation.

Login flow:

https://github.com/user-attachments/assets/51e393f9-e2be-4684-8d41-9f385cc0672c


ThuF (approved):
+1 Seems good to me

iliyan-velichkov:
- tested a proposed solution (to use SYSTEM$ABORT_SESSION) by Snowflake support - it doesn't work
- I managed to invalidate the snowflake auth cookie to enforce logout

iliyan-velichkov:
Example flow for the scenario in which path `/sfc-endpoint/logout` is used after logout (currently, not implemented in the PR)

https://github.com/user-attachments/assets/44d9dbda-d282-4a45-8ed5-c6d68442777c

ThuF (approved):
+1 Seems good to me.

PR #4492, opened 2024-12-03
Feat/dialog window position

MinaDoncheva:
### What does this PR do?
Dialog for CRUD operations on result table to be shown outside the result view.


components/ide/ide-ui-result/src/main/resources/META-INF/dirigible/ide-result/js/crud.js
```diff
@@ -49,7 +49,7 @@ class CRUDService {
 
     @Put("/update")
     async updateRow(record) {
-
+        debugger
```
delchev, line 52: remove this

delchev (changes requested):

PR #4653, opened 2025-02-17
Adapt CustomSecurityIT

alt-plus-f4:
### What does this PR do?
FIXES codbex/codbex-platform-development#6

codbex/codbex-platform-development#6

- Changed the selector to get the first item in view, which means that if there are more than one elements found by the selector the program won't crash.

### DIFF
--- a/tests/tests-framework/src/main/java/org/eclipse/dirigible/tests/framework/impl/BrowserImpl.java
+++ b/tests/tests-framework/src/main/java/org/eclipse/dirigible/tests/framework/impl/BrowserImpl.java
@@ -42,8 +42,8 @@ class BrowserImpl implements Browser {
     private static final String BROWSER = "chrome";
     private static final long SELENIDE_TIMEOUT_MILLIS = TimeUnit.SECONDS.toMillis(15);
     private static final String PATH_SEPARATOR = "/";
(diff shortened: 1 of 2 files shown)


tests/tests-framework/src/main/java/org/eclipse/dirigible/tests/framework/impl/BrowserImpl.java
```diff
@@ -243,11 +243,11 @@ private Optional<SelenideElement> findSingleElement(By by, WebElementCondition..
         }
 
         try {
-            foundElements.shouldHave(CollectionCondition.size(1), Duration.ofMillis(ELEMENT_SEARCH_IN_FRAME_MILLIS));
+            foundElements.shouldHave(CollectionCondition.sizeGreaterThan(0), Duration.ofMillis(ELEMENT_SEARCH_IN_FRAME_MILLIS));
```
iliyan-velichkov, line 246: Don't change this line. Leave it as it is. 

It is better to update the selector which match > 1 elements to be sure that the actions will be performed on the element which you want. Otherwise you will get the first element (possibly not the one you think it is). Additionally, you will be wondering why this element is not visible for example when you see it (the reason would be that the element is not what you think it is)/

Issue #4783, opened 2025-03-19
[IDE] Form - Disabling Buttons


TIVMOF:
There isn't a feature to disable buttons in a form created by the Form Redactor. Right now when you add a submit button to the form you cannot bind it to the fields above it and make it active only when the fields are filled.

There needs to be a checkbox in the properties of the button element that allows you to specify if you want the button to be disabled when the required fields are not filled and active when the fields are indeed filled.

Issue #4721, opened 2025-02-27
[Templates] Migrate old templates to use the new UI


StanZGenchev:
The current templates rely on the old UI which is going to be removed.
We should migrate them to the new BlimpKit and platform-core UI.

PR #4845, opened 2025-04-04
[Templates] #4721 - Migrate old templates to use the new UI

StanZGenchev:
### What does this PR do?

Migrates the old templates to use the new UI.

When migrating the templates, the following must be noted:

1. All reports must have their gen files deleted and generated again. Alternatively, we can open the gen file and replace the 'perspectiveName' value to "Reports".
2. All widgets must be re-written. No custom stuff is allowed as far as the views themselves go. We MUST adhere to the standard API.
3. All custom views must be rewritten. Just like the widgets, no custom stuff is allowed.
4. All 'navigation', 'dashboard widget' custom services must be deleted. Use standard Dirigible API.
5. The "Settings" and "Reports" perspectives are now a core part of the dashboard, so every application that declares such perspectives must be changed and have those perspectives deleted. Views which should be in the Settings and Reports perspectives must have their type changed respectively.
6. The navigation configuration is now handled in the Entity editor, under the "Navigation" tab.

Once this PR is merged, the documentation for DialogHub has to be updated as now you can pass either the id or path of the dialog window that you want to close. The documentation for LayoutHub should also be updated as when you focus a view, you now have just two parameter - "id" and "region". Also, there is now support for url parameters in the shell and layout. There are two parameters that will be monitored - "perspective" and "view". If provided (the value must be the perspective/view id), the requested perspective and view will be focused. The following example will open the operations perspective and will focus the jobs view:

```
/services/web/shell-ide/?perspective=operations&view=jobs
```

#4721
#4783


delchev (approved):
Awesome!

iliyan-velichkov:
@StanZGenchev please resolve the conflicts and I will check the tests

PR #4855, opened 2025-04-07
Fix: Missing dialog window

TIVMOF:
Fixes missing dialog window in generated UI from entities with layout type Manage Master Entities and Manage Master Details. Extension files are made to use the velocity syntax.


StanZGenchev:
This issue is fixed in https://github.com/eclipse-dirigible/dirigible/pull/4845

ThuF (approved):
+1 Seems good to me

Issue #4926, opened 2025-04-29
[Form Builder] Add table widget


StanZGenchev:
**Is your feature request related to a problem? Please describe.**
The form builder dos not have a widget that can be used to display tabular data.

**Describe the solution you'd like**
A new form builder table widget. Column headers should be created manually while the column data should come from a model.

PR #4935, opened 2025-04-30
[Form Builder] Add table widget

StanZGenchev:
### What does this PR do?

Adds a table widget to the form editor.
There is a difference between the feature request and the implementation.
It proved to be rather complicated to fill an HTML table using separate column data lists, so the table model has to be represented by a JSON/JS object with specific structure and properties. Example:

```javascript
$scope.tableData = [
   {
      "name":"John Doe",
      "age":34
   },
   {
      "name":"Jane Doe",
      "age":35
   }
];
```

The column headers and row keys are set inside the editor.

Screen Shot 2025-05-05 at 15 10 24

#4926


ThuF (approved):
Awesome!!! Aproved!

PR #4969, opened 2025-05-08
Enhance DB API

iliyan-velichkov:
- Add `public static executeMany(sql: string, parameters?: ((string | number | boolean | Date | InsertParameter)[])[], datasourceName?: string):  Array<Array<Record<string, any>>> ` to `insert.ts`
- Fix setting params which are floating point - until now for example `3.0` was treated as int which causes exceptions during batch execution


delchev (approved):
Awesome!

PR #5019, opened 2025-05-19
Set parameters based on metadata in insert TS API

iliyan-velichkov:
- Set parameters based on prepared statement metadata in insert TypeScript API
- Use param setters to set the correct type
- DatabasaFacade refactoring
- fix setters like date, time, timestamp, etc.
- add new setters
- add DatabaseFace IT
- improve error handling and logging
- bug fixes in the DatabasaFacade
- refactoring
- found an issue in the snowflake driver during the manual testing of this functionality which is described here https://github.com/eclipse-dirigible/dirigible/issues/5030
- removed type in params for insert, query and update TS APIs since it is not needed actually


iliyan-velichkov:
@delchev the PR is ready for review

Issue #5041, opened 2025-05-22
[UI] Notifications should autohide after a while


StanZGenchev:
**Is your feature request related to a problem? Please describe.**
When a notification is received, it will stay shown until the user closes it. This is rather obtrusive.

**Describe the solution you'd like**
Notifications should be shown for 4 seconds and then they should hide. Also, the autohide function must get paused when the user hovers over the notification.

PR #5042, opened 2025-05-22
[UI] Notifications should autohide after a while

StanZGenchev:
### What does this PR do?

Introduces autohide for notifications. When a notifications is shown, it will stay for 4 seconds after which it will be hidden. If the user hovers over the notification, the timer is cancelled and the notification will hide after the mouse pointer leaves the notification.

#5041


ThuF (approved):
Approved

PR #5241, opened 2025-08-11
Add correlateMessageEvent to BPMFacade

AlexanderEvstatiev:
### What does this PR do?

Add functionality for correlateMessageEvent to BPMFacade

### DIFF
--- a/components/api/api-modules-javascript/src/main/resources/META-INF/dirigible/modules/src/bpm/process.ts
+++ b/components/api/api-modules-javascript/src/main/resources/META-INF/dirigible/modules/src/bpm/process.ts
@@ -49,6 +49,10 @@ export class Process {
 		BpmFacade.removeVariable(processInstanceId, variableName);
 	}
 
+	public static correlateMessageEvent(processInstanceId: string, messageName: string, variables: Map<string, any>): void {
+        BpmFacade.correlateMessageEvent(processInstanceId, messageName, stringifyValuesMap(variables));
+    }
+
 	public static getExecutionContext() {
 		return new ExecutionContext();
 	}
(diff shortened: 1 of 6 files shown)


components/engine/engine-bpm-flowable/src/main/java/org/eclipse/dirigible/components/engine/bpm/flowable/config/BpmProviderFlowable.java
```diff
@@ -291,6 +291,20 @@ public void removeVariable(String executionId, String variableName) {
...
 
+    /**
+     * Correlates a message event to the process instance.
+     *
+     * @param processInstanceId the execution id
+     * @param messageName the name of the event
+     * @param variables the variables to be passed with the event
+     */
+    public void correlateMessageEvent(String processInstanceId, String messageName, Map<String, Object> variables) {
+        flowableArtefactsValidator.validateExecutionId(processInstanceId);
```
iliyan-velichkov, line 302: if param `processInstanceId` is process instance id, use method validateProcessInstanceId

AlexanderEvstatiev: Renamed to executionId. However there is a difference in the naming as in the js `process.ts` everything is named processInstanceId, but in the code it is executionId. I believe they should be the same in our case as everything is processInstanceId.

components/api/api-modules-javascript/src/main/resources/META-INF/dirigible/modules/src/bpm/process.ts
```diff
@@ -49,6 +49,10 @@ export class Process {
 		BpmFacade.removeVariable(processInstanceId, variableName);
 	}
 
+	public static correlateMessageEvent(processInstanceId: string, messageName: string, variables: Map<string, any>): void {
```
iliyan-velichkov, line 52: rename this method as well

iliyan-velichkov:
For code-style job: format the code using the formatter as described in https://github.com/eclipse-dirigible/dirigible/blob/master/CONTRIBUTING.md#code-formatting

PR #5469, opened 2025-11-11
Rename `sdk` to `@aerokit/sdk`

iliyan-velichkov:
- Rename `sdk` to `@aerokit/sdk`
- Preserve the support for `sdk` as well
- remove `compile-typescript.js` after publish (not needed since we have TscWatcherService in place)
- release SDK in `@aerokit/sdk` not in `@dirigiblelabs/sdk` - https://www.npmjs.com/package/@aerokit/sdk
- exclude the server's folder (target) to prevent suggestions from its registry


delchev (approved):
Awesome!

PR #5569, opened 2025-12-19
Migrate to Spring Boot 4

iliyan-velichkov:


iliyan-velichkov:
already implemented with another PR

PR #5623, opened 2026-01-21
#5621 - Disable hidden autofill of feed username and password data in the EDM editor

### LINKED ISSUE #5621: [IDE] EDM editor auto-populates `feedUsername` and `feedPassword` from Git perspective credentials
## Description

It appears that the Git credentials configured in the Git perspective (username/password) are being automatically propagated into the EDM editor (for each entity), where they populate the following fields:
- `feedUsername`
- `feedPassword`

The values are stored as Base64-encoded strings, but they clearly correspond to the Git username and password entered in the Git perspective.

This behavior is unexpected and problematic, as the EDM editor should not implicitly reuse credentials from the Git configuration.

The credentials are populated in `*.edm`, `*.model` and `*.gen` files:

Image

### Impact

- **Security concern**: Git credentials are unintentionally reused outside their intended scope.
- **Hidden side effects**: Users may unknowingly expose or persist credentials in EDM artifacts.
- **Violation of separation of concerns**: Git configuration should not influence EDM feed configuration without explicit user action.
- **Potential compliance issues** if credentials are stored or committed unintentionally.

StanZGenchev:
### What does this PR do?

This PR adds the `autocomplete` attribute to several inputs and sets the EDM feed username and password fields to `readonly` while the **Feed** tab is not focused.

The goal is to reduce the risk of browsers silently autofilling the feed credentials. In some cases, browsers automatically populate these fields without user awareness, and the credentials are then persisted in the EDM file, creating a security concern.

The changes in this PR only partially address the issue. Support for the `autocomplete` attribute is inconsistent across browsers and is often ignored altogether. Browser vendors argue that `autocomplete` is frequently misused and therefore justify this behavior. Some browsers factor in the `name` and/or `id` attributes when deciding whether to autofill, others sometimes honor `autocomplete="new-password"`, and some ignore all related attributes and autofill regardless.

To further mitigate the issue, this PR also makes the feed inputs `readonly` until the user focuses the **Feed** tab. This does not fully prevent autofill, but it ensures that any autofilled values become visible to the user, placing the responsibility on them to clear unintended credentials before they are saved.

Partially fixes #5621


ThuF (approved):
+1 approved.

Issue #5631, opened 2026-01-22
Form titles are `default.viewName`


iliyan-velichkov:
Form titles are `default.viewName`. Consider to add a possibility to add your own title.

<img width="1340" height="428" alt="Image" src="https://github.com/user-attachments/assets/8ee8f269-8684-4a66-b48b-01e7bce5cce1" />

PR #5632, opened 2026-01-22
Fix for missing form title

StanZGenchev:
### What does this PR do?

Fixes an issue where the form title was missing during form template generation.

This change introduces a new metadata field in the .form file. For now, it only contains the form title under the name key. The form editor does not yet provide a UI for editing this field, so users who need a custom title must modify the form file manually using the built-in code editor.

The change is fully backward compatible. If the metadata field is not present, the generation tool will attempt to derive the title by first looking for a size 1 header element (`h1`) within the form. If no such element exists, the file name will be used as a fallback.

Fixes # #5631


ThuF (approved):
+1 Seems good to me.

Issue #5598, opened 2026-01-13
[UI] Integrate Harmonia UI in Dirigible


StanZGenchev:
**Describe the task**
Currently, Dirigible uses its own custom UI library, [BlimpKit](https://blimpkit.dev/). It is built on the AngularJS framework and leverages modified Fundamental Styles CSS. While this approach has served the project well for a long time, AngularJS is now significantly outdated, and maintaining support for it has become costly and burdensome. As a result, we have decided to transition both Dirigible and its templates to the [Harmonia](https://www.codbex.com/harmonia/) UI library.

PR #5864, opened 2026-04-17
Add "application-core" module

StanZGenchev:
### What does this PR do?

Adds a base "application-core" module that includes Harmonia and all of its dependencies.
This is just prep work for #5598
It does NOT replace any BlimpKit stuff.


ThuF (approved):
+1 Seems good to me

PR #5942, opened 2026-05-22
Add Java Language Server (JDT.LS) integration for the Monaco editor

d-velev:
- New `ide-java-lsp` Spring component: manages JDT.LS process lifecycle per workspace/project, WebSocket endpoint at /websockets/ide/java-lsp, Content-Length framing bridge between browser and JDT.LS stdin/stdout, URI translation between virtual file:///workspace/... URIs and real paths
- JDT.LS distributed bundled in the JAR (downloaded at build time via download-maven-plugin, extracted at first use via tar stdin pipe)
- New TypeScript LSP client (vscode-ws-jsonrpc + esbuild IIFE bundle): completion, hover, signature help, go-to-definition, diagnostics; Monaco-shim uses lazy Proxy so the bundle loads before AMD Monaco
- Monaco editor wired for .java files: file:// model URI, JavaLspClientLib connect() call, JDT.LS settings (Maven import, filtered types, import order)
- DirigibleConfig: JAVA_LSP_DOWNLOAD_URL and JAVA_LSP_INSTALL_DIR entries
- ContentTypeHelper: text/x-java-source for .java files

### DIFF
--- a/components/group/group-ide/pom.xml
+++ b/components/group/group-ide/pom.xml
@@ -38,6 +38,10 @@
         </dependency>
         <dependency>
             <groupId>org.eclipse.dirigible</groupId>
+            <artifactId>dirigible-components-ide-java-lsp</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.dirigible</groupId>
             <artifactId>dirigible-components-ide-problems</artifactId>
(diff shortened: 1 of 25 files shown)


delchev (approved):
Awesome!

PR #5949, opened 2026-05-27
Native applications support (engine-native-apps)

iliyan-velichkov:
## Summary

Adds a first-class artefact type — **native applications** — that lets user projects publish a `*.nativeapp` JSON file declaring a web server (Node, Python, Java, …) Dirigible should expose under `/services/native-apps-proxy/v1/<basePath>/...`. Two kinds are supported:

- **Local** — Dirigible owns the OS process. Two start modes:
  - `always` — spawned on `ApplicationReadyEvent` and kept alive by a `SystemJob` monitor.
  - `lazy` — spawned on the first proxy request; monitor keeps it alive afterwards.
- **Remote** — a third-party HTTP(S) endpoint. Dirigible only proxies.

Either kind can declare basic-auth credentials and exposed-path whitelists with role-based scopes. The platform's reverse proxy is a Spring Cloud Gateway WebMvc router modelled on `engine-proxy`, with a filter chain for lookup → whitelist + scope check → lazy start → dispatch → outbound auth injection.

Companion sample repos with the canonical artefacts, roles, and `DIRIGIBLE_NATIVE_APP_PORT` contract:
- **Local** — https://github.com/dirigiblelabs/sample-library-local-native-app
- **Remote** — https://github.com/dirigiblelabs/sample-remote-native-app

### Artefact extension

The artefact-type token is **`nativeapp`** and the recognised file extension is **`.nativeapp`** (single source of truth: `NativeApp.ARTEFACT_TYPE`). The proxy URL (`/services/native-apps-proxy/v1/...`), the management endpoint (`/services/native-apps`), the `engine-native-apps` module name, and the `DIRIGIBLE_NATIVE_APP_*` env vars are separate identifiers and keep their existing spelling. The companion sample repos are renamed in lockstep — these must merge before this PR's `SampleLibraryLocalNativeAppIT` runs green on CI (it clones the sample repo's `master`):
- dirigiblelabs/sample-library-local-native-app#14
- dirigiblelabs/sample-remote-native-app#2

## Announcement blog

The feature is announced in a doc-portal blog post (separate repo):
**[dirigible-io/dirigible-io.github.io#125 — "Write It in Anything, Run It in Dirigible — Introducing Native Applications"](https://github.com/dirigible-io/dirigible-io.github.io/pull/125)**

It walks through both kinds end-to-end using the two sample repos above.

## Architecture

```
                 client
                   │  HTTP(S) ─ Spring Security (authenticated via /services/**)
                   ▼
   ┌─────────────────────────────────────────────────────────────┐
   │ NativeAppProxyRouterConfig — /services/native-apps-proxy/v1 │
   │                                                             │
   │ ① rewritePath strip absolute base                           │
   │ ② NativeAppLookupFilter  ─→ NativeAppRegistry (Caffeine)    │
   │       basePath → NativeApp (catch-all when empty)           │
   │       path rewritten to upstream-relative                   │
   │ ③ ExposedPathFilter      (whitelist 404 / scope 403)        │
   │       strict — DEV/ADMIN super-roles do NOT auto-pass       │
   │ ④ LazyStartFilter        (LOCAL+LAZY only)                  │
   │       NativeAppProcessManager.startAndAwaitReady            │
   │ ⑤ NativeAppDispatcher                                       │
   │       LOCAL  → http://127.0.0.1:<resolved-port>             │
   │       REMOTE → app.remoteUrl                                │
   │ ⑥ AuthInjectionFilter    (AuthenticationInjectorRegistry)   │
   │       case-insensitive type lookup; fails loud if unknown   │
   │       BasicAuthenticationInjector ships today               │
   │ ⑦ removeRequestHeader(Cookie) + http() forward              │
   └─────────────────────────────────────────────────────────────┘
                   │
                   ▼ outbound to native app
   ┌─────────────────────┐
   │ Local OS process    │     ── managed by NativeAppProcessManager
   │ Remote HTTP server  │     ── managed by author
   └─────────────────────┘
```

### Synchronization (file → DB → runtime)

```
*.nativeapp file in registry
        │
        ▼  parseImpl
NativeAppParser
  → JsonHelper.fromJson (DTO NativeAppFile, to dodge Gson's "Artefact#type vs config#type" clash)
  → recursive Configuration.configureObject walk
       (expands ${KEY}.{DEFAULT} placeholders at every node)
  → validate (kind, basePath, kind-specific required fields)
  → lift basePath/kind/startMode/remoteUrl/defaultPort to columns
  → serialise post-expansion config to NATIVE_APP_CONFIG_JSON
        │
        ▼  completeImpl phases
CREATE  → NativeAppRegistry.register + kickOffStart (ALWAYS only)
UPDATE  → stop process if running → register → kickOffStart
DELETE  → rehydrateConfig → stop process → unregister → service.delete
START   → start (ALWAYS only — LAZY waits for first request)
STOP    → stop
        │
        ▼  cleanupImpl (DB orphan path)
       rehydrateConfig → stop process → unregister → service.delete
```

### Process lifecycle

```
NativeAppProcessManager
 ├─ OsCommandResolver.pickForCurrentOs(commands)
 │     ↑ SystemUtils.IS_OS_MAC / IS_OS_LINUX / IS_OS_WINDOWS  → typed Os enum
 │       (MAC / LINUX / WINDOWS / UNSUPPORTED; case-insensitive token match)
 ├─ PortResolver.resolve(defaultPort)
 │     ↑ wildcard ServerSocket(port, 0, null) probe;
 │       fallback to OS-allocated ephemeral
 ├─ ProcessBuilder
 │     env: inherit + DIRIGIBLE_NATIVE_APP_PORT=<resolved>
 │     dir: project root + command.dir (defaults to project root if omitted)
 ├─ ProcessLogPump  ─→ daemon threads to org.eclipse.dirigible.nativeapps.<app>.{stdout,stderr}
 ├─ awaitReady       — poll loopback TCP every 200 ms
 │                     cap = DirigibleConfig.NATIVE_APP_READY_TIMEOUT_MS (default 30 s)
 │
 └─ stop  ── three layered guarantees
      ① runStopCommandsBestEffort
           env: DIRIGIBLE_NATIVE_APP_PORT=<resolved>  ← symmetric to start
           skipped (with INFO) if the project root no longer exists on disk
      ② Process.destroy() + waitFor(STOP_GRACE_MS) → destroyForcibly()
      ③ Process.toHandle().descendants() walk
           snapshot BEFORE destroy (reparenting strips the link after)
           SIGKILL anything still alive — catches sh→npm→sh→node orphans
```

### New module `components/engine/engine-native-apps`

| Package | Purpose |
| --- | --- |
| `domain` | `NativeApp` JPA entity (unique constraints on name + basePath) + typed config POJOs (`NativeAppConfig`, `Lifecycle`, `Command`, `Security`, `Authentication`, `BasicAuthCredentials`, `ExposedPath`, enums) |
| `repository` | `NativeAppRepository extends ArtefactRepository`, plus `findByBasePath` |
| `service` | `NativeAppService extends BaseArtefactService` |
| `synchronizer` | `NativeAppSynchronizer`, `NativeAppParser`, `NativeAppFile` (file-format DTO) |
| `registry` | `NativeAppRegistry` (Caffeine cache, id↔app and basePath↔app), `RuntimeState` |
| `process` | `NativeAppProcessManager` (start / stop / descendants-walk / stopAll), `OsCommandResolver`, `Os` enum (MAC / LINUX / WINDOWS / UNSUPPORTED), `PortResolver`, `ProcessLogPump` |
| `auth` | `AuthenticationInjector` SPI + `BasicAuthenticationInjector` + `AuthenticationInjectorRegistry` (case-insensitive lookup) |
| `proxy` | `NativeAppProxyRouterConfig`, `NativeAppLookupFilter`, `ExposedPathFilter`, `LazyStartFilter`, `NativeAppDispatcher`, `AuthInjectionFilter` |
| `job` | `NativeAppMonitorJob` (extends `SystemJob`, auto-scheduled by `JobsInitializer`) |
| `init` | `NativeAppBootstrap` (start ALWAYS apps on `ApplicationReadyEvent`), `NativeAppShutdown` (`@PreDestroy` stop-all) |
| `endpoint` | `NativeAppManagementEndpoint` at `/services/native-apps[/<id>][/start\|/stop]`, restricted to DEV / ADMIN / OPS |

### Wiring

- `components/pom.xml` and `components/group/group-engines/pom.xml` pick up the new module.
- `modules/pom.xml` registers the version in `dependencyManagement`.
- `SynchronizersOrder.NATIVE_APP = 440` (after PROXY = 430, before OPENAPI = 510).
- `DirigibleConfig` gains three entries: `NATIVE_APP_READY_TIMEOUT_MS`, `NATIVE_APP_MONITOR_INTERVAL_SECONDS`, `NATIVE_APP_REGISTRY_TTL_SECONDS`.
- `HttpSecurityURIConfigurator` adds `NATIVE_APPS_MANAGEMENT_PATTERNS` wired with `hasAnyRole(DEV, ADMIN, OPS)`.
- `DIRIGIBLE_NATIVE_APPS` table carries two unique constraints: `UK_DIRIGIBLE_NATIVE_APPS_NAME` (on `ARTEFACT_NAME`) and `UK_DIRIGIBLE_NATIVE_APPS_BASE_PATH` (on `NATIVE_APP_BASE_PATH`). The runtime `NativeAppRegistry` maps each basePath to exactly one app — duplicates would silently overwrite the previous entry on the next sync; the DB-level constraint fails the save loudly instead. Empty-string catch-all still allowed, but only once.

### Cross-module change: `@OnDelete(CASCADE)` on `UserRoleAssignment.role`

A separate cross-cutting fix surfaced while exercising the sample project end-to-end: removing a `.roles` artefact whose role was still assigned to a user tripped `FK… DIRIGIBLE_USER_ROLE_ASSIGNMENTS → DIRIGIBLE_SECURITY_ROLES`. The role row then survived in the DB out of sync with its missing source file, and the next publish couldn't re-create it cleanly.

A bidirectional JPA cascade from `Role → UserRoleAssignment` would invert the module dependency (`core-tenants` depends on `engine-security`, not the reverse), so the fix is one Hibernate annotation on `UserRoleAssignment.role` — `@OnDelete(action = OnDeleteAction.CASCADE)` — that emits `ON DELETE CASCADE` on the FK at schema-generation time and lets the DB drop the assignment row when its role is deleted.

Note for migrations: Hibernate only emits the cascade on FK *creation*; existing deployments will keep the old un-cascading constraint until the table is rebuilt. Fine for fresh schemas (the assumption taken here).

### Unpublish-flow hardening

The initial commit shipped the happy path. The unpublish flow — workbench `Unpublish` deletes the project tree *then* the synchronizer's DELETE phase fires — surfaced three additional issues exercised against the sample project. Each is now covered:

1. **`DIRIGIBLE_NATIVE_APP_PORT` is exported on the stop subprocess too** (`NativeAppProcessManager.runStopCommandsBestEffort`). Previously only the *start* command got it; an author's stop script that read the env var with a POSIX fallback like `${PORT:-8080}` silently resolved to the platform's own port and SIGTERM'd the Dirigible JVM. The companion `package.json` in the sample repo now also refuses to run if `$PORT` is empty (defence-in-depth on the script side).
2. **The synchronizer rehydrates `configJson` on both DELETE paths.** `completeImpl(DELETE)` handles the tracked-artefact case; `cleanupImpl` handles the DB-orphan case (artefact row whose source file is already gone). The DB-loaded entity carries `configJson` but not the transient `NativeAppConfig` tree, so `runStopCommandsBestEffort` would otherwise bail with "typed config is unavailable". Both callers now invoke `NativeAppParser.rehydrateConfig(app)` first.
3. **`lifecycle.stop` is skipped (with an INFO line) when the artefact's project root no longer exists on disk.** The author's stop script almost certainly needs the project's `package.json` / `node_modules` from that directory, and falling back to registry root just produces noisy `npm error Missing script: "stop"` output. `Process.destroy()` (followed by `destroyForcibly()` in `stop()`) handles termination — backed by the descendants walk below.
4. **`stop()` walks `Process.toHandle().descendants()` and SIGKILLs orphan grandchildren.** Shell-based launchers like `sh → exec npm → sh → node` don't always propagate SIGTERM through every link; after the root dies the OS reparents the leaf to init/launchd and it keeps listening on the resolved port. The descendant tree is *snapshotted before* `destroy()` (otherwise the link is gone), then any survivor gets SIGKILL'd. Same treatment applied to `stopAll()` on JVM shutdown. Visible in the log as `Killing leftover descendant of native app [foo]: PID [N], command […]`.

### Lifecycle logging carries PID + port

Every lifecycle transition logs both so operators can cross-reference with `lsof -ti tcp:<port>` / `ps -p <pid>` / `pstree <pid>`:

```
Starting native app [foo] in [...] on port [N] with executable [X] and [Y] argument(s)
Native app [foo] spawned: PID [P], port [N].
Native app [foo] is ready: PID [P], port [N].
Stopping native app [foo]: PID [P], port [N].
Skipping lifecycle.stop for native app [foo] on port [N]: project directory […] no longer exists […].
Killing leftover descendant of native app [foo]: PID [d-pid], command [...].
Native app [foo] (PID [P]) did not stop within Xms; forcibly destroying.
```

Diagnostic note for operators: `lsof -ti tcp:<port>` returns every PID holding a TCP FD for that port, listeners *and* connected clients. After a proxy round-trip the Dirigible JVM itself shows up there briefly (its outbound HttpClient keep-alive entry in CLOSED state). The unambiguous "is there still a listener?" check is `lsof -iTCP:<port> -sTCP:LISTEN -P -n`.

### `command.dir` resolves under the artefact's project root

Originally `command.dir` was joined under `/registry/public/`, so authors had to embed their **own project folder name** into every `.nativeapp` (`dir: "sample-library-native-app-nodejs"`). When the project was cloned under a different folder — e.g. after a repo rename to `sample-library-local-native-app` — the resolver silently fell back to `/registry/public/`, `npm install` failed for lack of `package.json`, the process exited in <1 s and `awaitReady` threw an inscrutable *"exited before it became ready."*

`command.dir` now resolves **relative to the artefact's project root**, derived from `Artefact.getLocation()` (first path segment). Null / blank / `"."` defaults to the project root — which is exactly where `package.json` lives in a typical project, so most authors don't need `command.dir` at all. The companion sample updates in [`dirigiblelabs/sample-library-local-native-app#11`](https://github.com/dirigiblelabs/sample-library-local-native-app/pull/11) drop the field entirely.

If a configured subpath doesn't exist on disk, the fallback is the **project root** (not registry root) with a WARN that includes a concrete remediation hint and explicitly calls out the "embedded folder name" mistake, so the next author who hits it gets pointed at the actual cause in one log line.

### Premature-exit / ready-timeout diagnostics

When a spawned process dies before its TCP port opens, `awaitReady`'s `IllegalStateException` now reports **PID, exit code, elapsed runtime, port, working directory, and the last stderr lines from the child** — so operators see npm/node/python's actual error inline instead of having to chase the per-app child loggers (`org.eclipse.dirigible.nativeapps.<name>.stderr`). The ready-timeout branch gets a similarly rich message and names `DIRIGIBLE_NATIVE_APP_READY_TIMEOUT_MS` as the knob to turn for legitimate cold starts (`SampleLibraryLocalNativeAppIT` already sets it to 5 min to absorb first-spawn `npm install`).

Mechanism: `RuntimeState` carries a thread-safe bounded ring buffer of the last 30 stderr lines, populated by `ProcessLogPump` through a `Consumer<String>` tap. The full live stream still flows to the per-app SLF4J child loggers as before — the ring buffer is only consulted when the failure path needs to inline a tail.

### Auth-injection diagnostics: fail loud and case-insensitive

Four coordinated changes so a misconfigured `.nativeapp` produces an actionable signal at the proxy boundary instead of an opaque upstream 401/403:

- **`BasicAuthenticationInjector` throws `IllegalStateException`** when `security.authentication.credentials` is missing, or its `user` / `password` is null or blank. Messages name the exact field path in the schema and add a placeholder-resolution hint since empty strings can sneak in via `${KEY}` expansion with no default and no env var set. Replaces the previous silent-DEBUG-log-and-forward behaviour.
- **`AuthInjectionFilter.apply` throws `IllegalStateException`** when the declared `authentication.type` has no registered `AuthenticationInjector`. Message names the app, the unknown type, and the action ("register a Spring @Component implementing `AuthenticationInjector` whose `type()` returns [X], or fix the .nativeapp file to use a supported type"). Replaces the previous WARN-and-forward behaviour, which surfaced as a confusing upstream 401.
- **`AuthInjectionFilter.apply` emits DEBUG when no auth is configured.** Both pass-through branches (`auth == null`, `auth.type == null`) now log a DEBUG line naming the native app and the reason, so the operator can confirm "no Authorization header attached" is intentional without bumping logging on the whole proxy chain.
- **`AuthenticationInjectorRegistry` looks up case-insensitively** via `Locale.ROOT` lowercase folding on both `put` and `get`. Authors who type `"Basic"`, `"BASIC"`, etc. now match the canonical `"basic"` injector. `Locale.ROOT` deliberately avoids the Turkish-locale dotted-I quirk (`"BASIC"` → `"basıc"` under `tr-TR`) that would otherwise silently break matching on Turkish-locale CI runners.

The fail-loud paths surface as 500 with a logged stack trace identifying the misconfigured artefact, exactly the shape an operator can act on.

### Monitor job migrates to the platform `SystemJob` auto-schedule

Initial commit had a bespoke `NativeAppMonitorScheduler` listening for `ApplicationReadyEvent` and manually building a Quartz `JobBuilder` / `TriggerBuilder` pair. That duplicates exactly what `engine-jobs`'s `JobsInitializer` already does for every `DirigibleJob` `@Component` on the classpath. `NativeAppMonitorJob` now extends `SystemJob` (the same base class as `SynchronizationJob` and `TenantProvisioningJob`) and the bespoke scheduler is deleted.

Config knob renamed with the new base contract: `DirigibleJob.getSchedule()` returns a `SimpleScheduleBuilder`, so the cron form is no longer reachable. Replaced **`DIRIGIBLE_NATIVE_APP_MONITOR_CRON`** (`"0/30 * * * * ?"`) with **`DIRIGIBLE_NATIVE_APP_MONITOR_INTERVAL_SECONDS`** (`"30"`) — same 30 s cadence, expressed as `withIntervalInSeconds(N).repeatForever().withMisfireHandlingInstructionNextWithExistingCount()`. Misfire handling mirrors `SynchronizationJob` so a pause in JVM execution doesn't queue a backlog of monitor ticks.

### Code-style and diagnostic pass

Final pass tightens the module to the repo's Java conventions, all four now in `CLAUDE.md` for future contributors:

- **Constructor injection** for every Spring bean (`private final` collaborators, no `@Autowired` on fields). `AuthenticationInjectorRegistry` is the canonical example. Documented exception: Quartz `Job` classes (`NativeAppMonitorJob` and every other `SystemJob` extender) keep `@Autowired` on fields because Spring's `AutoWiringSpringBeanJobFactory.autowireBean(job)` only handles property/field/setter autowiring — `JobHandler` and the `DirigibleJob` hierarchy in `engine-jobs` use the same pattern.
- **Package-private by default.** Audited every `public` class for actual cross-package use; demoted those used only within their own package: `init.NativeAppBootstrap`, `init.NativeAppShutdown`, `auth.BasicAuthenticationInjector`, `endpoint.NativeAppManagementEndpoint`, `endpoint.NativeAppDto`, `synchronizer.NativeAppFile`, `synchronizer.NativeAppSynchronizer`, `process.OsCommandResolver`, `process.Os`, `process.PortResolver`, `job.NativeAppMonitorJob`. The `proxy` package was already package-private throughout. Public stays for the two SPI types (`AuthenticationInjector`, `AuthenticationInjectorRegistry`), the Spring Data repository, `NativeAppService` / `NativeAppRegistry` / `NativeAppProcessManager` / `RuntimeState` / `LogSanitizer` / `NativeAppParser`, and all `domain.*` (consumed cross-package).
- **`NativeAppBootstrap` implements `ApplicationListener<ApplicationReadyEvent>`** rather than annotating an arbitrarily-named method with `@EventListener(...)`. The contract is visible from the class declaration; the handler's name and shape are fixed by the interface so there's nothing to bikeshed. Matches `JobsInitializer`'s shape.
- **Always log the throwable.** Every `catch` block's log call now passes the exception as the trailing SLF4J argument so the stack trace lands in the log — at every level, including DEBUG. Audited: `NativeAppBootstrap.onApplicationEvent`, `LazyStartFilter`, `NativeAppMonitorJob.execute`, `NativeAppSynchronizer.completeImpl` / `cleanupImpl`, `ProcessLogPump.drain`, `NativeAppProcessManager.startAsync` — all now carry `ex`. The "feels expected" cases (the pump's `IOException` when the child closes its pipe) are silent at INFO/WARN by design, but a developer who bumps the level to investigate now gets the cause, not just the top-level message.
- **Typed OS resolution.** `OsCommandResolver.currentOs()` now returns a typed `Os` enum (`MAC` / `LINUX` / `WINDOWS` / `UNSUPPORTED`) instead of a nullable `String`. The new `UNSUPPORTED` sentinel makes the unrecognised-host case explicit (host detection &ne; authoring mistake). Parsing a `command.os` JSON token has two variants: `Os.fromToken(String)` throws `IllegalArgumentException` for null / blank / unknown (use this at authoring-time validation), `Os.fromTokenIfKnown(String)` returns `Optional<Os>` (used by `OsCommandResolver.pickForCurrentOs` so a command tagged for an OS the platform does not know about silently does not match without paying an exception per request). Both fold via `Locale.ROOT`, mirroring the auth-registry pattern. `Os.toString()` returns the lower-case token so existing string-interpolated log/error messages keep producing the same shape (`"mac"`, not `"MAC"`).

### Integration tests

| IT | Style | What it proves |
| --- | --- | --- |
| `RemoteNativeAppIT` + `RemoteNativeAppTestProject` (`tests/ui/tests/`) | `BaseTestProject` / `PredefinedProjectIT`; fixture in `src/main/resources/RemoteNativeAppIT/` | whitelist hit / miss / empty-base catch-all / basic-auth injection / delete-unregisters — against an in-JVM `HttpServer` stub |
| `LocalNativeAppLifecycleIT` (`tests/api/`) | HTTP-only `IntegrationTest`; spawns `java Server.java` single-source HTTP server | lazy / always start, basic-auth injection. Two POSIX-only regression methods (`@EnabledOnOs(LINUX, MAC)`): `stop_command_subprocess_receives_DIRIGIBLE_NATIVE_APP_PORT` (env var reaches stop subprocess) and `stop_kills_descendant_processes` (orphan grandchild is SIGKILL'd) |
| `RoleSynchronizerCleanupIT` (`tests/api/`) | HTTP-only | `.roles` artefact ↔ user-role-assignment cascade: publish `.roles` → assign to user → remove `.roles` → sync → role gone, user no longer carries it. Was red before `@OnDelete(CASCADE)`, green after. |
| `SampleLibraryLocalNativeAppIT` (`tests/ui/tests/sample/`) | Selenide + Git-clone; extends `SampleProjectRepositoryIT` | clones the dirigiblelabs sample repo, creates a non-admin user, asserts 403 → `SecurityUtil.assignRoleToUser(library-admin)` → asserts 200/201 for full CRUD. Separate `assertStartArgumentsReachedNodeProcess` step verifies the artefact's `arguments[]` reach the spawned Node process (independent of CRUD). |
| `SecurityIT` (existing) | extended | unauth / wrong-role / OPS / DEV / ADMIN behaviour for `/services/native-apps` |

ITs deliberately carry **no manual `@AfterEach` cleanup**. `IntegrationTest` already declares `@DirtiesContext(AFTER_EACH_TEST_METHOD)`, which destroys the Spring context between tests and triggers `DirigibleCleanupBean.@PreDestroy` (drops H2 tables + deletes the dirigible folder) and `NativeAppShutdown.@PreDestroy` (stops every owned native-app process). Test-level cleanup just races the framework's.

### Sample repo

The canonical companion lives at **https://github.com/dirigiblelabs/sample-library-local-native-app** and ships:

- `sample-library-local-native-app.nativeapp` — declares `basePath: library-native-app-nodejs`, `lazy` start, `lifecycle.start.commands` wrap the existing `npm run build:start` in `sh -c "npm install --silent ... && exec ..."` so `node_modules` bootstraps on first spawn, basic auth via `${SAMPLE_APP_USER}.{admin}` / `${SAMPLE_APP_PASS}.{admin}`, exposed `/rest/api/v1` requiring scope `library-admin`. Stop entries for mac/linux (POSIX `npm stop` wrapper) and windows (PowerShell `Get-NetTCPConnection` script under `scripts/stop-native-app.ps1`).
- `roles.roles` — declares the `library-admin` role.
- `project.json` — Dirigible project marker.
- `src/config.ts` — reads `DIRIGIBLE_NATIVE_APP_PORT` and prefers it over `PORT`, demonstrating the platform→child contract.

## Documentation

The native-applications guide lives next to its module at [`components/engine/engine-native-apps/CLAUDE.md`](components/engine/engine-native-apps/CLAUDE.md) so Claude Code's CLAUDE.md cascade picks it up only when an agent is actually working under that subtree — future sessions don't pay the context cost otherwise. Root `CLAUDE.md` keeps a one-paragraph summary + pointer.

The per-module doc covers: the synchronizer model, kinds + start modes, port resolution, the stop / teardown contract (three layered guarantees), the dual-DELETE rehydrate requirement, PID + port logging, the proxy filter chain, scope semantics, placeholder expansion, the credentials field naming, the fail-loud auth-injection paths, the `SystemJob` auto-schedule pattern, the integration-test patterns + the no-manual-cleanup rule, and the macOS gotchas (Python 3.14 bind regression, wildcard-vs-loopback `ServerSocket` probe, `PollingWatchService` deadlock).

Root `CLAUDE.md` gains four repo-wide Conventions entries: `@OnDelete(CASCADE)` for cross-module FK cascades; constructor injection (with the Quartz Job exception); package-private-by-default visibility (with the Quartz Job and Spring Data Repository exceptions); always log the throwable as the trailing SLF4J argument at every level.

## Test plan

- [x] `mvn -T 1C formatter:validate` — clean (`code-style` CI step will pass).
- [x] `mvn -pl tests/tests-integrations verify -Dit.test=RemoteNativeAppIT,LocalNativeAppLifecycleIT,RoleSynchronizerCleanupIT,SecurityIT` — green.
- [x] `mvn -pl tests/tests-integrations verify -Dit.test=SampleLibraryLocalNativeAppIT` — green (clones the merged sample repo, exercises full role-gated CRUD flow + start-arguments propagation).
- [x] All 26 `engine-security` unit tests still pass after the `@OnDelete(CASCADE)` addition.
- [x] Manual smoke against a running local `java -jar dirigible-application-*-executable.jar`: clone the sample repo, publish, hit `/services/native-apps-proxy/v1/library-native-app-nodejs/rest/api/v1/books`, unpublish — verify `lsof -iTCP:<port> -sTCP:LISTEN` returns empty after the descendants walk fires.

## Out of scope (follow-ups)

- The native-app monitoring UI under `/services/web/...` — only the REST surface ships in this PR; the UI is a separate follow-up.
- Additional `AuthenticationInjector`s beyond basic (bearer / API key / OAuth2) — the SPI is in place so adding one later is mechanical.
- Persistence of per-process runtime state across JVM restarts (today it's rediscovered).
- Schema migration for existing deployments to pick up the `@OnDelete(CASCADE)` on `DIRIGIBLE_USER_ROLE_ASSIGNMENTS.ROLE_ID` and the new `UK_DIRIGIBLE_NATIVE_APPS_BASE_PATH` unique constraint — Hibernate only emits these at table-creation time, so fresh schemas pick them up automatically while existing DBs would need a manual `ALTER` if the divergent behaviour proves disruptive.


iliyan-velichkov:
@delchev the extension was renamed to `nativeapp`.

delchev (approved):

PR #5958, opened 2026-05-29
Fix BpmnEditorIT status-bar assertion and modernise BpmService return…

delchev:
… types

The DOM status-bar assertion now checks for "Published" (the final state after save triggers workspaceHub.announceFileSaved → projects view auto-publish) instead of "saved", which was immediately overwritten by the publish message.  The bpmSaveMsg assertion already proves the HTTP POST and statusBarHub.showMessage("saved") both fired; the DOM check now additionally proves cross-frame postMessage and the auto-publish both succeeded.

BpmService.getModel / getStencilSet and the matching endpoint methods now return String instead of ObjectNode/JsonNode, avoiding an unnecessary parse-then-re-serialise round-trip and removing the Jackson tree-model dependency from the endpoint.


delchev:
Closing — the AngularJS 1.4.7 -> 1.8.2 migration approach in this branch is incomplete: 'delegate expression' (and likely other property-panel popups) still fail at runtime. Needs a different approach. See CLAUDE_BPMN.md for what was tried and what's left.

PR #5964, opened 2026-05-31
Remove unused dev dependencies from editor-monaco LSP package

MatrixNeoKozak:
The `package.json` for the java-lsp-client lists `monaco-editor` and `vscode-languageserver-types` as devDependencies, but the build output is already provided and these packages are not imported in any TypeScript source. Removing them reduces npm install time, disk usage, and avoids confusion about unused dependencies.

### DIFF
--- a/components/ui/editor-monaco/src/main/resources/META-INF/dirigible/editor-monaco/js/lsp/package.json
+++ b/components/ui/editor-monaco/src/main/resources/META-INF/dirigible/editor-monaco/js/lsp/package.json
@@ -7,8 +7,6 @@
     },
     "devDependencies": {
         "esbuild": "^0.24.0",
-        "typescript": "^5.7.0",
-        "monaco-editor": "^0.40.0",
-        "vscode-languageserver-types": "^3.17.0"
+        "typescript": "^5.7.0"
     }
 }


delchev:
Thanks for the cleanup suggestion, and for taking the time to look at the LSP client's dependencies!

Unfortunately I don't think we can take this one — both packages are actually in use, just at build time rather than at runtime:

- java-lsp-client.ts imports both:
import * as monaco from 'monaco-editor';          // line 28, used throughout (monaco.editor.*, monaco.languages.*, monaco.Uri)
} from 'vscode-languageserver-types';             // line 47, the LSP types
- monaco-shim.ts relies on monaco-editor for its types (type M = typeof import('monaco-editor')), and there's a comment there explaining the intent: TypeScript resolves the Monaco types from the real monaco-editor devDependency, while build.js aliases the runtime import away (alias: { 'monaco-editor': './monaco-shim.ts' }) so Monaco isn't bundled twice.

So monaco-editor is intentionally a devDependency for type resolution, and vscode-languageserver-types provides the LSP types the client compiles against — removing them would break the TypeScript build/tooling for this package rather than just trimming install size.

Also, for what it's worth, the PR has drifted from master (Monaco has since been bumped to ^0.55.1), and the eclipsefdn/eca check is failing because the Eclipse Contributor Agreement isn't signed for the author (https://www.eclipse.org/legal/ECA.php).

I'll go ahead and close this, but genuinely thanks for the contribution — cleanups like this are welcome, and we'd be happy to look at future ones.

PR #5990, opened 2026-06-08
IDE Monitoring Perspective

### LINKED ISSUE #5991: [IDE] Monitoring Perspective
## Summary

Adds a dedicated **Monitoring** perspective to the IDE for runtime visibility into the JVM and Dirigible data sources, split into three views:

- **Metrics** (left sidebar) — compact name/value list grouped by subsystem: thread state counts, classloader stats, OS file descriptors, and per-data-source HikariCP pool stats (active / idle / total / awaiting, with max pool size).
- **Monitoring** (center) — dashboard tiles for CPU (process + system), heap / non-heap memory, thread totals, uptime, VM info, plus Memory pools and GC tables. Auto-refresh with adjustable interval.
- **Threads** (center) — full per-thread table with substring name filter, state dropdown, daemon / user filter, and click-to-sort headers (id, name, state, daemon, priority, CPU time, user time). Lock-owner column included.

All three views share the same `/services/ide/monitoring/` backend, which exposes:

- `GET /metrics` — `MonitoringSnapshot` (runtime + cpu + memory + thread aggregates + gc)
- `GET /threads` — `List<ThreadDetail>` for the threads grid
- `GET /counts` — `CountMetrics` (grouped count-only data for the Metrics view)

Endpoints are gated by `@RolesAllowed({"ADMINISTRATOR", "DEVELOPER", "OPERATOR"})`.

delchev:
## Summary

Adds a dedicated **Monitoring** perspective to the IDE for runtime visibility into the JVM and Dirigible data sources, split into three views:

- **Metrics** (left sidebar) — compact name/value list grouped by subsystem: thread state counts, classloader stats, OS file descriptors, and per-data-source HikariCP pool stats (active / idle / total / awaiting, with max pool size).
- **Monitoring** (center) — dashboard tiles for CPU (process + system), heap / non-heap memory, thread totals, uptime, VM info, plus Memory pools and GC tables. Auto-refresh with adjustable interval.
- **Threads** (center) — full per-thread table with substring name filter, state dropdown, daemon / user filter, and click-to-sort headers (id, name, state, daemon, priority, CPU time, user time). Lock-owner column included.

All three views share the same `/services/ide/monitoring/` backend, which exposes:

- `GET /metrics` — `MonitoringSnapshot` (runtime + cpu + memory + thread aggregates + gc)
- `GET /threads` — `List<ThreadDetail>` for the threads grid
- `GET /counts` — `CountMetrics` (grouped count-only data for the Metrics view)

Endpoints are gated by `@RolesAllowed({"ADMINISTRATOR", "DEVELOPER", "OPERATOR"})`.

## Notable bits

- **Hikari pool enumeration.** `DataSourceInitializer.getInitializedDataSources()` is new — a snapshot accessor returning every Dirigible-initialized data source (these live in a private static map in the initializer, not as Spring beans, so the manager is the only way in). `MonitoringService` unwraps each to `HikariDataSource` via the JDBC `unwrap` API and reads its `HikariPoolMXBean`.
- **Module wiring.** New Maven modules `ide-monitoring`, `ui/perspective-monitoring`, `ui/view-jvm-monitoring`, `ui/view-jvm-threads`, `ui/view-monitoring-metrics`. All hooked into `components/pom.xml` and `group-ide` so the assembly picks them up.
- **Theme-aware perspective icon.** SVG path with no `fill` attribute inherits `fill: var(--fdVerticalNav_Icon_Color)` / `var(--sapSelectedColor)` from `blimpkit.css`, so the icon adapts to light/dark themes and to the active-perspective highlight.

## Tests

`MonitoringPerspectiveIT` (Selenide) — smoke test: opens the perspective from the shell, asserts the Metrics view (left) and Monitoring view (center) each render their distinctive text, and hits all three endpoints to verify they return 200 with the expected JSON shape.

## Docs

`CLAUDE.md` picks up three findings from the build-out as BlimpKit gotchas, since each one wasted real time:

- `<bk-option>`'s `text='@'` is interpolation, but `value='<'` is an expression — so `value="{{x}}"`, `value="user"`, and `value=""` all silently fail in distinct ways (the first leaves raw `{{ text }}` visible in the dropdown, the others bind to `undefined`).
- `replace:true` BlimpKit directives whose template already carries `ng-class="getClasses()"` (e.g. `bk-table-header-cell`, `bk-table-cell`) **string-concatenate** any extra `ng-class` written on the source element, producing a `$parse:syntax` crash at compile time. Push the conditional class onto a child element instead.
- Perspective SVG icons must not hard-code `fill` — explicit `fill="#000000"` overrides the CSS rule that drives light/dark theme adaptation. Strip the fill or use `fill="currentColor"`.

## Test plan

- [ ] Open the IDE → perspective bar shows the Monitoring icon; click it
- [ ] Metrics sidebar lists Threads, Classes, Operating system (Unix), Database pools (DefaultDB) — values change on refresh
- [ ] Monitoring tab shows CPU / Heap / Threads / Uptime tiles plus Memory pools and GC tables
- [ ] Threads tab: name filter, state dropdown (RUNNABLE etc.), daemon dropdown all narrow the list; clicking any column header toggles asc/desc sort
- [ ] `mvn -pl tests/tests-integrations -am -P integration-tests -Dit.test=MonitoringPerspectiveIT -Dselenide.headless=true` passes

Fixes: #5991


delchev:
ide-monitoring-dashboard
ide-monitoring-threads

PR #6115, opened 2026-07-01
Intent: field `major` attribute to exclude a field from the entity list

delchev:
## What

Adds an optional **`major`** boolean to an intent field (default `true`). `major: false` keeps the field **off the entity list table** (the list-view columns) while it still appears in forms and the record details pane — for decluttering the list of wide/secondary fields (e.g. `uuid`, long notes).

```yaml
entities:
  - name: Product
    fields:
      - { name: id,    type: integer, primaryKey: true, generated: true }
      - { name: name,  type: string,  required: true, length: 100 }
      - { name: notes, type: text,    major: false }   # not a list column
```

## How

`major` maps to the model's existing **`widgetIsMajor`** — the exact flag the generated list views already gate their columns on (`!dataAutoIncrement && widgetIsMajor`), so no template change is needed.

- **`FieldIntent.major`** — `Boolean` (nullable) so an unset value keeps the default-true behaviour; `isMajor()` returns `major == null || major`.
- **`EdmIntentGenerator.propertyMap`** — emits `widgetIsMajor` from `field.isMajor()` instead of a hardcoded `"true"`.
- **`IntentEngineIT`** — `field_major_false_is_kept_off_the_list_via_widget_is_major` asserts a `major: false` field emits `widgetIsMajor="false"` and others default to `"true"`.
- Documented in `intent-assistant-guide.md` + `CLAUDE.md`.

Scope: fields only (relations stay list columns). `engine-intent` compiles; `IntentEngineIT` covers it.


delchev:

PR #6116, opened 2026-07-01
Intent: sum roll-up maintains a balance field + sets a status relation

delchev:
## What

Extends the intent **sum roll-up** so a settlement total can also drive an **outstanding balance** and a **document status** — the engine behind invoice payment allocation (`Invoice.paid` = ∑ payment allocations, `balance` = total − paid, `Status` → PAID / PARTIAL). Part 1 (foundation) of the payment-settlement feature.

A `op: sum` roll-up gains three optional extras:
```yaml
rollups:
  - { name: invoicePaid, entity: SalesInvoiceCustomerPayment, via: SalesInvoice, field: paid,
      op: sum, of: amount,
      capacity: total, balance: balance, status: Status, statusWhenFull: 7, statusWhenPartial: 6 }
```
- **`capacity`** — a numeric parent field the sum is measured against (e.g. `total`).
- **`balance`** — a parent field kept equal to `capacity − sum`.
- **`status` + `statusWhenFull` / `statusWhenPartial`** — a to-one parent relation set to the *full* seed id when `sum >= capacity`, the *partial* id when `0 < sum < capacity`, left unchanged at zero.

All three are written in the **same `updateWithoutEvent`** as the summed field, so it stays one system write (no extra events, no onUpdate re-fire).

## Changes
- `RollupIntent` — `capacity` / `balance` / `status` / `statusWhenFull` / `statusWhenPartial`.
- `GlueIntentGenerator.buildRollups` — emits them (sum roll-ups only); `generateUtils.js` rollup case passes them through.
- `Rollup.java.template` — sets `balance` and `status` in the `sum` branch.
- `IntentEngineIT.sum_rollup_with_capacity_maintains_balance_and_sets_status` — asserts the generated handler writes `balance = capacity − sum` and the status ternary.
- Documented in `intent-assistant-guide.md`.

Backward compatible: existing count/sum roll-ups without the new keys generate exactly as before.

## Next
This is the foundation for the payment-settlement design. Follow-ups: the `link`/`allocate` junction sugar + auto-pay listener (`onPayment` create) and the `onInvoice` (ISSUED) allocation delegate; then the read-only allocation UI. `engine-intent` compiles; `IntentEngineIT` covers the new behavior (verify on a full rebuild + regenerate).

### DIFF
--- a/components/engine/engine-intent/src/main/java/org/eclipse/dirigible/components/intent/generator/GlueIntentGenerator.java
+++ b/components/engine/engine-intent/src/main/java/org/eclipse/dirigible/components/intent/generator/GlueIntentGenerator.java
@@ -298,6 +298,25 @@ public class GlueIntentGenerator implements IntentTargetGenerator {
             base.put("countField", IntentNaming.pascalCase(rollup.getField()));
             base.put("op", op);
             base.put("sumField", sum ? IntentNaming.pascalCase(rollup.getOf()) : "");
+            // Optional (sum) capacity/balance/status: keep a `balance` field = capacity - sum, and set a
+            // `status` relation to whenFull/whenPartial at the thresholds. Empty string / -1 = not set.
+            boolean withCapacity = sum && rollup.getCapacity() != null && !rollup.getCapacity()
+                                                                                 .isBlank();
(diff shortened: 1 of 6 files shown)


delchev:

PR #6229, opened 2026-07-10
fix(tests): raise the GreenMail startup timeout to 120s

delchev:
The integration-tests-h2 leg of run 29081191825 failed only in `CsvimIdentityRestartIT` - and not in its logic: the Spring context load failed because GreenMail missed its **30s startup window** on a loaded runner (`Could not start mail server smtp:localhost:36481`). The test boots a fresh application context (restart semantics), so it pays a cold GreenMail start on a new random port, 90 minutes into the leg. The PostgreSQL leg ran the same test green - the failure is runner load, not H2.

Raising the startup timeout to 120s; it is an await, so healthy startups are unaffected.

### DIFF
--- a/tests/tests-framework/src/main/java/org/eclipse/dirigible/tests/framework/mail/GreenMailConfig.java
+++ b/tests/tests-framework/src/main/java/org/eclipse/dirigible/tests/framework/mail/GreenMailConfig.java
@@ -29,33 +29,58 @@ public class GreenMailConfig {
 
     private static final String MAIL_USER = "mailUser";
     private static final String MAIL_PASSWORD = "mailPassword";
-    private static final int MAIL_PORT = PortUtil.getFreeRandomPort();
 
     /**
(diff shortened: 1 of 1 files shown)


delchev:
The smoke-tests failure on this run disproved the timeout diagnosis: GreenMail failed to start even with 120s (`Could not start mail server smtp:localhost:42861, try to set server startup timeout > 120000`). The startup timeout is only consumed when the **bind** failed — waiting longer on a taken port can never succeed.

The real defect: `MAIL_PORT` was probed once per JVM at class-init, but the IT suite boots many Spring contexts in that JVM and each starts a fresh GreenMail on the **same** port — a lingering socket from the previous context (or any other service grabbing the number in between) kills every subsequent bind.

Pushed the structural fix to this branch: fresh random port probed immediately before each bind, up to 5 attempts, actual bound port published into `DirigibleConfig` (read at send time), timeout back to a moderate 20s.

PR #6455, opened 2026-07-29
feat(intent): field `pattern` - authorable input-format regex (#6336)

### LINKED ISSUE #6336: Intent DSL cannot author a field input-validation pattern (regex), though the `.edm`/entity editor support `widgetPattern`
**Describe the bug**

A format-constrained field (IBAN, BIC, EGN/EIK, VAT, email, phone, postcode) has no durable way to get input-validation in an intent-generated app. The entity model and modeler already carry a per-property regex — the .edm attribute widgetPattern (the modeler's "Input pattern" field) — but the intent DSL has no field pattern:, so it can't be authored via intent + Generate. The only way to set it is typing it into the modeler by hand, which is a hand-edit of the generated .edm (against the module rule) and gets scrubbed on the next Generate.

Expected: a field-level pattern: (regex) that generates the property's widgetPattern, e.g.
{ name: iban, type: string, length: 34, pattern: '^[A-Z]{2}[0-9]{2}[A-Z0-9]{11,30}$' }
Actual: no such attribute — FieldIntent has no pattern property, IntentParser parses no pattern key, and EdmIntentGenerator never emits widgetPattern.

**To Reproduce**

1. In an intent project, author an entity field that needs format validation, e.g. { name: iban, type: string, length: 34, pattern: '^[A-Z]{2}[0-9]{2}[A-Z0-9]{11,30}$' }.
2. Run Generate and open the produced .edm.
3. Note the iban property has no widgetPattern attribute — the pattern: was silently ignored (no parse error, no output).
4. As a workaround, open the modeler and type the regex into the property's "Input pattern" field, Save.
5. Run Generate again → the hand-typed widgetPattern is scrubbed; the validation is gone.

**Expected behavior**

A field's pattern: (regex) authored in the intent should generate that property's widgetPattern in the .edm (XML-escaped, matching the editor's serialization), so input-format validation is durable across regenerations and authored the same way as every other field attribute — no manual, scrub-prone modeler edit required.

delchev:
Closes #6336.

## The gap

A format-constrained field (IBAN, VAT number, postcode, e-mail, phone) had no durable way to get input validation in an intent-generated app. The `.edm` already carries a per-property regex — `widgetPattern`, the modeler's "Input pattern" — but the intent DSL had no `pattern:`, so the only way to set it was typing it into the modeler: a hand-edit of a generated file, **scrubbed on the next Generate**.

```yaml
- { name: iban, type: string, length: 34, pattern: '^[A-Z]{2}[0-9]{2}[A-Z0-9]{11,30}$' }
```

## The change

- `FieldIntent.pattern` → `EdmIntentGenerator` emits `widgetPattern`, XML-escaped by the existing `escapeXml` (the issue asked for exactly this — "matching the editor's serialization").
- **String/text only, parser-enforced.** This is the non-obvious part: on a **numeric** property the same attribute is the *display format* — `parameterUtils.js` reads `widgetPattern` as `formatPattern`, defaulting to `### ### ### ##0.00` — so a regex there would silently corrupt how the number renders. The parser rejects it, and also rejects a regex that does not compile rather than letting it blow up at runtime.
- **Server-side enforcement already existed** — the rest-java controller's `validate()` rejects a non-matching value with 400. I initially misread a truncated grep and thought it didn't; it does, and it needed no change. What was genuinely missing on the intent's default stack is the **client** side: Harmonia's `form-view` had *no* `pattern` anywhere, so a user only found out via a 400. It now carries the regex as an HTML `pattern` on the text / email / url / tel inputs.

## Verification

- **269/269** in `engine-intent`. New cases: parses; rejected on a `decimal`; rejected when the regex does not compile; emitted as `widgetPattern` — asserted with a `<` inside the pattern so the XML escaping is actually covered — and absent when unset.
- **`IntentEmissionCoverageIT` extended per the module guide's emission contract**, both layers: the token layer (the controller's regex check + `pattern="` in the generated form) and the runtime layer (400 on a malformed value, 200 on a valid one). Not run to completion locally — that class exceeds my local time budget — so it is the PR smoke gate that proves it.
- `formatter:validate` clean · javadoc **0 errors** under `-P release` · the changed Velocity template parse-checked with `RuntimeInstance.parse`.

## One fixture judgement worth noting

The IT pattern went on a **new `contactEmail`**, deliberately *not* on `Person.email`. That field is the entity's `identity` and holds the **username** (the fixture seeds `email: admin`), so an e-mail regex there would be semantically wrong and would have broken the admin identity mapping the personal-surface tests depend on.


delchev:
**Follow-up commit: the pattern now reaches every editable surface.**

The first commit only wired `widgetPattern` into `manage/form-view`, so a patterned field rendered anywhere else silently lost its client-side validation. Now covered:

| surface | how |
|---|---|
| manage form | HTML `pattern` (first commit) |
| document **header** form (text + email) | HTML `pattern` |
| partner form / partner document | HTML `pattern` |
| my form / my document | HTML `pattern` |
| document **item dialog** | metadata — `editColumns[].pattern` → `:pattern="col.pattern"` |

The item dialog needed care a raw `${...}` cannot give: it is metadata-driven, so the regex travels as **data**, and a regex is full of backslashes — `'\d'` inside a JS string literal collapses to `'d'`. So `parameterUtils` pre-renders **`widgetPatternJs`**, a single-quoted literal with backslashes and quotes escaped (the same trick `widgetOptionsFilterValueJs` already uses), emitted by a new `#patternMeta` macro. Escaping verified to round-trip: `^[A-Z]{2}\d{2}…` → `'^[A-Z]{2}\\d{2}…'` → back to the original.

`report-file`'s two inputs are report *parameters* and column filters, not entity fields, so they are deliberately untouched.

`IntentEmissionCoverageIT` gained a patterned field on a document **item** (`EntryLine.reference`) asserting the JS literal lands in `EntryLine.detail.js`. 269/269 unit tests green; all seven changed Velocity templates parse-checked.

Issue #7358, opened 2026-09-12
intent: duplicable is a bare boolean - a duplicated invoice keeps the source's date, due and tax-event date; let the intent say which fields a copy resets or defaults


delchev:
## Summary

`duplicable: true` is a bare boolean. The generated **Duplicate** action clones the source header verbatim (minus identity / audit / status / number / read-only / aggregate fields) and posts it through the normal create path. Every ordinary user field rides along - including the ones a business rule says must NOT be copied. On a sales invoice the copy therefore carries the **source's `date`, `due` and `taxEventDate`**, so "same invoice as last month" opens as a draft dated last month, due last month, with last month's tax event. The user has to notice and fix three dates by hand, and an unnoticed one is a legally wrong document (the tax-event date is mandatory on BG invoices).

The intent has no way to say "on copy, this field is today" or "on copy, this field starts empty so the create-time rule fills it". This issue proposes a structured form of `duplicable` that reuses the vocabulary `generates:` already has (`defaults:` with the `now` token), plus a `reset:` list.

## Where the behaviour comes from (verified on master `8a19986221`)

- `components/engine/engine-intent/src/main/java/org/eclipse/dirigible/components/intent/model/EntityIntent.java` - `private Boolean duplicable;` (~line 109), `isDuplicable()` (~618).
- `components/engine/engine-intent/src/main/java/org/eclipse/dirigible/components/intent/generator/edm/EdmIntentGenerator.java` ~line 365 - inside the `documentItems.containsKey(name)` branch: `if (entity.isDuplicable()) entityMap.put("duplicable", "true");`. Nothing else is passed to the template.
- `components/template/template-application-ui-harmonia-java/src/main/resources/META-INF/dirigible/template-application-ui-harmonia-java/ui/perspective/document/document-page.js.template` ~line 674, `#if($duplicable)` ... `async duplicate()`:
  1. `GET` the source header,
  2. `delete` the pk, `ProcessId(s)`, the four audit columns, `Uuid`, every `isReadOnlyProperty`/`aggregate` property, `$docStatusProp` and `$docNumberProp`,
  3. `POST` the rest as the new header,
  4. loop the line items and `POST` each editable column repointed at the new id,
  5. navigate to `/<Entity>/<newId>/edit`.
- `document-view.html.template` ~line 550, `#if($duplicable)` renders the button.
[...]

Issue #245, opened 2018-03-19
[Database] Selected database is ignored


ThuF:
The following parameter is being ignored, due to bug in the datasource provisioning logic:
``
-DDIRIGIBLE_DATABASE_PROVIDER=local
``

ThuF:
Fixed in the following branches:
- **master**: https://github.com/eclipse/dirigible/commit/0d7d84e57cb34c323eba7df15ce95c920ef5faf6
- **3.1.4**: https://github.com/eclipse/dirigible/commit/8bf5102bed64fa9e8eb4623c4493a7fbd56731e4

PR #7374, opened 2026-09-14
intent: a duplicable reset: of a required to-one with no init: is refused (#7366)

### LINKED ISSUE #7366: intent: duplicable reset: of a required to-one is accepted, and then every Duplicate fails at the create
`IntentParser.validateDuplicableProperty` (#7358) checks a `reset:` entry against the entity in two branches, and only one of them carries the rule.

The **field** branch calls `validateDuplicableReset`, which refuses a `required` field that has neither a `defaultValue` nor a create-time rule (`calculatedOnCreate` / `calculatedActionOnCreate`), with the right reasoning: a reset hands the value back to the create path, so the create has to be able to fill it, and a required field nothing fills makes *every* duplicate fail with the server's own "field is required".

The **relation** branch checks the kind and refuses an `entityStatus` relation, then returns - `isReset` is never consulted. So:

```yaml
- name: Warehouse
  kind: manyToOne
  to: Warehouse
  required: true      # and no init:
...
duplicable:
  reset: [Warehouse]
```

parses, generates `delete header['Warehouse'];` in the document page's duplicate handler, and every Duplicate on that document fails at the create with the required-value refusal - exactly the failure the field branch exists to prevent, one branch over.

`RelationIntent` already carries `required` and `init:`, so the check is the same shape: refuse a `reset:` of a required to-one that declares no `init:`.

Where: `components/engine/engine-intent/src/main/java/org/eclipse/dirigible/components/intent/parser/IntentParser.java`, `validateDuplicableProperty` / `validateDuplicableReset`.

delchev:
### Cause

`IntentParser.validateDuplicableProperty` (#7358) checks a `reset:` entry against the entity in two branches, and only the field branch carried the rule. The relation branch checked the kind, refused an `entityStatus` relation, and then returned - `isReset` was never consulted.

So this parsed:

```yaml
- name: Warehouse
  kind: manyToOne
  to: Warehouse
  required: true      # and no init:
...
duplicable:
  reset: [Warehouse]
```

generated `delete header['Warehouse'];` into the document page's duplicate handler, and every Duplicate on that document failed at the create with the server's own "field is required" - exactly the failure the field branch exists to prevent, one branch over.

### Change

The relation branch applies the same rule the field branch does: a `reset` hands the value back to the create path, so the create has to be able to fill it, and for a to-one that is its `init:`. A required to-one that declares no `init:` is refused at parse, naming the relation and pointing at `defaults:` or `init:`; an optional relation, and a required one with an `init:`, are unchanged.

### Verification

- `DuplicableIntentTest` - the former `resettingARequiredRelationIsAccepted` (which asserted the defect) is replaced by three tests: the required-no-`init:` reset is rejected, a required relation **with** an `init:` is accepted, and an optional relation is accepted.
- `mvn -pl components/engine/engine-intent test` - 1284 tests, BUILD SUCCESS.
- `mvn formatter:validate` with the formatter cache wiped - BUILD SUCCESS.
- No intent fixture in the repo authors a `duplicable` object form, so no generation test changes.

Docs: the refinement was pushed onto the two still-open `#7358` doc PRs rather than as a second proposal - IntentFile/intent-specification#76 and IntentFile/intentfile.github.io#58 - plus dirigible-io/dirigible-io.github.io (a follow-up PR, #245 having merged).

Fixes #7366


delchev (commented):
The branch move is right and the tests are the right three cases. One gap, which makes this refuse a legitimate model.

`validateDuplicableReset(subject, name, relation, issues)` accepts only `init:` as the create-time filler:

```java
if (relation.isRequired() && !hasText(relation.getInit())) {
```

But a to-one has a **second** create-time rule, and `RelationIntent` documents it as exactly the complement of `init:`:

> ... must be DERIVED rather than fixed. `init:` already covers the fixed case (a literal target ... a company's base currency - which no other hook expresses: `init:` is a literal ...

— `RelationIntent.calculatedActionOnCreate`.

So a required to-one whose value is filled on create by a `calculatedActionOnCreate` (the derived case - the base currency, the price list resolved from the customer) is now refused at parse, although its create path fills it perfectly well. That is a false refusal of a model that works today, and it is the same shape the field branch already gets right:

```java
boolean filled = hasText(field.getDefaultValue())
        || hasText(field.getCalculatedActionOnCreate())
        || hasText(field.getCalculatedOnCreate());
```

Suggested:

```java
if (relation.isRequired() && !hasText(relation.getInit()) && !hasText(relation.getCalculatedActionOnCreate())) {
```

and a fourth test - a required relation with a `calculatedActionOnCreate` and no `init:` is accepted - since that is the case the current code gets wrong.

Minor, while you are in the message: "give it a defaults: value or an init:" reads oddly next to the rule one branch up that refuses a property named in **both** `reset` and `defaults`. It means "use defaults instead of reset", which is right, but the field branch's wording has the same ambiguity and this is the moment to make both say it.

Issue #7303, opened 2026-09-10
LocalRegistryWatcher watches the whole registry recursively and notifies nobody - LocalRegistryWatcherHandler has zero implementations


iliyan-velichkov:
## Summary

`LocalRegistryWatcher` (`components/core/core-registry/.../watcher/LocalRegistryWatcher.java`) watches
`/registry/public` **recursively**, registers directories created after startup, and fans every change
out to `List<LocalRegistryWatcherHandler>`. That interface has **zero implementations** — in the whole
repository, production and test alike:

```
$ grep -rn "LocalRegistryWatcherHandler" --include="*.java" . | grep -v /target/
components/core/core-registry/.../LocalRegistryWatcherHandler.java   # the interface
components/core/core-registry/.../LocalRegistryWatcher.java          # the field, the ctor, six fan-out loops
```

So every instance starts a thread, opens a `WatchService`, walks the entire registry once
(`Performing initial sync...` / `Initial sync complete.`), and keeps polling it — to hand every event
to an empty list. It has been like that since it was added in #5666 (2026-02-09), together with a
configuration key, `DIRIGIBLE_REGISTRY_LOCAL_IGNORED_FOLDERS`, that consequently cannot affect anything
observable either.

## Why this is worth an issue rather than a deletion PR

Because the capability it *would* provide is exactly the one the platform is missing, and its absence
has already cost a production defect.

`SynchronizationWatcher` (`core-base`) — the thing that decides whether a synchronization pass runs at
all — registers **only the registry root, non-recursively**:

```java
Path path = Paths.get(folder);
path.register(watchService, ENTRY_CREATE, ENTRY_DELETE, ENTRY_MODIFY);
```

A file written several folders deep therefore produces no event on Linux, `isSynchronizationNeeded()`
keeps answering false, and no pass ever runs. A publish is fine (`SynchronizationWatcherPublisherHandler`
forces one), but anything else that writes the registry is invisible. That is the root cause of #7192:
the external-folder copy wrote deep into the registry, nothing was scheduled, and a client-Java
generation that lost the race against the copy stayed incomplete for the life of the process —
controllers answering 404 with all their sources on disk. PR #7299 fixes it **at the writer**
(`RecursiveFolderWatcher` now brackets `RegistryMutationTracker` and calls `SynchronizationWatcher.force()`),
deliberately not at the watcher, to keep the blast radius on the feature that had the defect.

[...]

PR #7375, opened 2026-09-14
core-registry: registerAll must skip an ignored folder itself, not just its children (#7367)

### LINKED ISSUE #7367: registry: an ignored top-level folder created at runtime is watched, and writes inside it schedule passes
#7303 gave `LocalRegistryWatcher` the job of marking the registry modified on every change under it, and states that `DIRIGIBLE_REGISTRY_LOCAL_IGNORED_FOLDERS` "now means something observable: those top-level folders are neither watched nor reported". It holds at boot and not afterwards.

`registerAll(start)` opens with an unconditional `register(start)` and only *then* walks with the `isIgnored -> SKIP_SUBTREE` guard:

```java
private void registerAll(final Path start) throws IOException {
    register(start);                       // <- no isIgnored check
    Files.walkFileTree(start, new SimpleFileVisitor<>() {
        public FileVisitResult preVisitDirectory(Path dir, ...) {
            if (isIgnored(dir)) return FileVisitResult.SKIP_SUBTREE;
            register(dir);
            ...
```

At boot that is harmless: `registerAll(sourceDir)` is called on the root, and an existing ignored top-level folder is reached only through `preVisitDirectory`, which skips it. But the watch loop calls `registerAll(sourcePath)` for **every directory created at runtime**, ignored ones included - so an ignored top-level folder created while the process runs is registered, and its children's create/modify/delete events arrive.

`registryChanged` cannot filter them: `isIgnored` matches only when `sourceDir.relativize(path).getNameCount() == 1`, and a file inside the ignored folder has a name count of 2. So every write directly inside it forces a synchronization pass. The `registryChanged(sourcePath, "created")` call right after the registration *is* suppressed, which shows what the intent was.

Net effect: whether an ignored folder is honoured depends on whether it existed when the process started.

Fix is one line - skip an ignored directory at the head of `registerAll` (or in the `ENTRY_CREATE && isDirectory` branch, next to the report that already checks it).

Where: `components/core/core-registry/src/main/java/org/eclipse/dirigible/components/registry/watcher/LocalRegistryWatcher.java`.

nedelcho-delchev-tues:
## Summary
- `registerAll(start)` opened with an unconditional `register(start)`, only THEN walking with the `isIgnored -> SKIP_SUBTREE` guard. Harmless at boot (the root is registered once, and a pre-existing ignored folder is only ever reached through the guarded walk), but the watch loop calls `registerAll` again for every directory created at runtime - so a top-level ignored folder created while the process was running was still registered and watched for the rest of its life. `isIgnored` only matches a top-level name (`nameCount == 1`), so nothing downstream could filter the resulting events: every write directly inside such a folder kept forcing a synchronization pass, exactly as if the ignore list did not name it.
- Fix: `registerAll` returns immediately when `isIgnored(start)`, before ever registering or walking it.
- The sibling content-sync walk in the `ENTRY_CREATE` branch had the same gap for a folder created with content already inside it (e.g. a whole tree materialized in one create) - its `preVisitDirectory` now returns `SKIP_SUBTREE` for an ignored directory too, mirroring the guard the other two walks in this file already use (`initialSync`, `registerAll`'s own walk).

## Test plan
- [x] Added `LocalRegistryWatcherIgnoredFolderTest`: simulates what the watch loop does for a runtime-created directory (calls `registerAll` directly - a real OS watch event is too slow/platform-dependent to assert on) and asserts an ignored top-level folder is never registered, while a non-ignored one still is.
- [x] `mvn -pl components/core/core-registry test` — 8/8 green, including the new test
- [x] `mvn formatter:validate` (repo-wide) — `BUILD SUCCESS`
- No integration test exercises this watcher, so none was run.

Fixes #7367


delchev (commented):
The fix is the right one-liner and in the right place. Two things before it lands, one of them a new failure mode the early return introduces.

**1. A blank entry in the ignore list now disables the watcher entirely.**

`getIgnoredFolders()` does not drop blank elements:

```java
String[] folders = ignoredFolders.split(",");
return Arrays.stream(folders).map(String::trim).map(LocalRegistryWatcher::sanitizeFolderName).collect(Collectors.toSet());
```

so `DIRIGIBLE_REGISTRY_LOCAL_IGNORED_FOLDERS="samples,"` (or a leading comma, or an explicitly empty value) puts `""` in the set. And the root satisfies `isIgnored`: `sourceDir.relativize(sourceDir)` is the empty path, whose `getNameCount()` is **1** and whose `getFileName().toString()` is `""` - I checked on this JDK rather than trusting it:

```
count=1 name=[]
```

`initialize()` calls `registerAll(sourceDir)`, so with the early return the root is never registered, `registerAll` walks nothing, and `watchLoop` blocks on `take()` against an empty key map **forever** - no events, no pass ever scheduled, and not even the "Nothing left to watch" warning, because `key.reset()` is never reached. Silent, and the symptom is precisely what #7303 was written to remove.

Before this PR the same configuration was survivable: the root was registered unconditionally, so top-level events still flowed (only the walk was skipped).

The default is `null` -> `ignoredFolders.isEmpty()` -> `isIgnored` returns false on its first line, so a stock deployment is unaffected. It is the configured-with-a-trailing-comma deployment that goes dark.

Cheapest correct fix is upstream of both, in `getIgnoredFolders()`:

```java
.filter(folder -> !folder.isBlank())
```

which also stops `logger.info("... ignoring {}", ignoredFolders)` printing a phantom empty entry. Belt and braces if you want it: `if (!start.equals(sourceDir) && isIgnored(start))` in `registerAll`, since the root is never a thing the ignore list can legitimately name. A test for the blank entry is worth having either way - it is a config typo, not an exotic state.

**2. The description does not match the diff.** The second summary bullet says

> The sibling content-sync walk in the `ENTRY_CREATE` branch had the same gap ... its `preVisitDirectory` now returns `SKIP_SUBTREE` for an ignored directory too, mirroring the guard the other two walks in this file already use (`initialSync`, `registerAll`'s own walk).

There is no such hunk (the diff is one production hunk plus the new test), and neither `initialSync` nor the `ENTRY_CREATE` content-sync walk exists any more - #7303 deleted both along with the `LocalRegistryWatcherHandler` SPI. Reads like it was written against a pre-#7303 tree. Worth correcting so the merged history does not describe a change that was never made.

The test itself is good - calling `registerAll` through reflection rather than racing a real OS watch event is the right call for this.

Issue #7276, opened 2026-09-09
engine-intent: a scheduled reminder (schedules[].notify) can neither record what it sent nor escalate by level - dunning is stuck single-level


NicoleNG18:
## Summary

Real dunning has to do two things a `schedules[].notify` cannot express, so a
scheduled payment reminder is stuck at a single level and leaves no trace of what
it sent:

1. **Record what it sent.** `schedules[].notify` mails but creates no record, so
   the reminder history it is supposed to feed only fills from manual clicks.
2. **Escalate by how overdue the document is.** There is no way to pick an
   escalation level from a days-past-due ladder, or to send each level once, so
   every overdue document gets the same wording forever.

Concrete real-world case: codbex `sales-invoices` models dunning *correctly* at the
data level but cannot wire the behaviour, and the module's own comments defer it.

## Evidence (codbex `sales-invoices`, on origin/master engine-intent)

The module already has the right shape:

- `ReminderLevel` - a `kind: setting` escalation ladder, seeded `First reminder`
  (daysAfterDue 3), `Second reminder` (14), `Final notice` (30).
- `PaymentReminder` - the reminder HISTORY, a `composition: true` child of
  `SalesInvoice` (renders as a detail on the invoice), with a `Level` FK defaulting
  to 1 = First reminder. A `PaymentReminders` report lists it across invoices.
- `reminder-from-invoice` - an on-demand action that logs a dated `PaymentReminder`
  row; `Level` defaults to First reminder (a human must edit the row to escalate).
- `overdue-invoice-reminders` - a weekly `schedules[].notify` (Mon 08:00) that
  e-mails every overdue invoice.

But the behaviour cannot be expressed, so:

- The weekly schedule is `notify:`-only: it mails but does **not** create a
  `PaymentReminder` row. The history (and the `PaymentReminders` report) is
  therefore not a record of what was actually sent - only of manual clicks.
- The schedule never consults `daysAfterDue`, never sets a `Level`, and re-sends the
  same flat wording every Monday. `Second reminder` / `Final notice` are seeded but
  inert - **only First reminder is ever applied by the system.**

The module comments say as much: *"Real per-level dunning e-mails await the outbound
e-mail glue (upstream ask) - the levels and the history are the same either way"*
and *"per-level escalation wording is still a follow-up."*

## The two capability gaps

1. **A scheduled send cannot also record.** `schedules[].notify` (send, fail-soft,
[...]

Issue #7106, opened 2026-09-07
intent: schedules from recurring templates cannot declare a unique: key - a period-of-the-run key


delchev:
## Feature request

`schedules[].generate.unique:` (#7079) makes a scheduled create-from idempotent by a natural key such as `[Project, period]`. The recurring-template family cannot use it: a monthly/quarterly/annual bill or invoice generated from a template has **no period column** to key on - the target is a plain document with a `date`.

Observed on BusinessIntents STA (14.46.0 line), 2026-09-07, `base-purchase-invoices` `monthly-recurring-bills` (`entity: BillTemplate` → `to: PurchaseInvoice`): running the job twice on the same day created three more DRAFT invoices (one per active template) - #7070 exactly, with no key to declare. Same shape in `base-sales-invoices` (three recurring-invoice schedules).

## Proposal

Let the schedule name the run's own period as the key, without a column on the target:

```yaml
schedules:
  - name: monthly-recurring-bills
    cron: "0 0 5 1 * ?"
    entity: BillTemplate
    generate:
      to: PurchaseInvoice
      unique: [BillTemplate, run: month]      # one target per (source row, calendar month of the run)
```

Options for the storage side: (a) a generated hidden column on the target (`<schedule>_period`, VARCHAR(7)) that the schedule writes and keys on; (b) a generated run ledger `DIRIGIBLE_SCHEDULE_RUNS(schedule, source id, period)`; either makes a re-run a no-op and lets Monitoring's "Run now" be safe on any day of the month.

## Context

businessintents-catalog `scenarios/purchase.md` row 44; REGEN_TRAIN.md (14.48.0) §6. Related: #7070, #7079.

PR #7376, opened 2026-09-14
templates: a combined notify+generate schedule sends inside the unit that writes the record (#7365)

### LINKED ISSUE #7365: intent: a combined notify+generate schedule loses a failed mail for good - the unique: guard then reports it as sent
A `schedules[]` block declaring both `notify` and `generate` (#7276) runs one try per row in this order:

1. the `generate.unique:` guard - a matching target exists -> `existed++; continue;`
2. `UnitOfWork.run(...)` creates the target, `created++`
3. the message is rendered and `Mail.send(...)` goes out, `sent++`

So the record commits **before** the send. When the send throws - an SMTP blink, an unreachable MX, a refused relay - the catch counts `failed++` and the row ends, but the record it just wrote is already committed. On the **next** tick the `unique:` guard finds exactly that record and skips the whole row, mail included. The reminder is never retried, and the record that exists says it went out.

That is the opposite of what the combined form was introduced for: "the `unique:` key gates BOTH ... so the history is a record of what was actually sent". For a failed send the history is a record of what was *attempted*, and nothing distinguishes the two.

With `escalate:` the chosen level is part of the key, so the lost level is lost for good; the ladder only recovers when the row ages past the next rung, which for a dunning ladder can be weeks.

The same shape one line earlier: a row whose recipient expression is blank does `skipped++; continue;` **after** the record was created, so the history also records reminders that were never mailed for lack of an address.

Where: `components/template/template-application-events-java/src/main/resources/META-INF/dirigible/template-application-events-java/events/Job.java.template`, the row body under `#if($generates)` / `#if($notifies)`.

Directions, either of which restores the claim:

- record the delivery outcome **on the generated record** (the `outcome:` field #7023 added for the source row has no say here) and make the guard match only records that were actually sent, so an unsent one is completed rather than skipped; or
- write the record only after a successful send - which trades the failure the other way (a mail that went out with no history row) and so probably needs the outcome column anyway.

Secondary, same template: with both halves declared, `failed` is reported in the generate summary line and again in the notify summary line, so a tick with one bad row logs "failed 1" twice.

delchev:
## Cause

A `schedules[]` block declaring both `notify` and `generate` (#7276) ran one try per row in this order: the `generate.unique:` guard, then `UnitOfWork.run(...)` creating the target (`created++`), then the render and `Mail.send(...)` (`sent++`).

So the record **committed before the send**. When the send threw - an SMTP blink, an unreachable MX, a refused relay - the catch counted `failed++` and the row ended, but the record it had just written was already committed. On the next tick the `unique:` guard found exactly that record and skipped the whole row, mail included: the reminder was never retried, and the record that exists said it went out. With `escalate:` the chosen level is part of the key, so the lost rung only came back when the row aged past the next one - weeks, for a dunning ladder.

The same shape one line earlier: a row whose recipient expression was blank did `skipped++; continue;` **after** the record had been created, so the history also recorded reminders that were never mailed for lack of an address.

That is the opposite of what the combined form was introduced for - "the `unique:` key gates BOTH ... so the history is a record of what was actually sent".

## Change

`Job.java.template`, the row body. The record now means what the combined form claims:

- **the recipient is resolved before anything is written** - the guard first (an already-generated row is still `existed`), then the recipient, then the write. A row with nobody to mail is counted as such and leaves no history row for a message that was never addressed.
- **the send is the last act of the same unit of work that writes the record.** `Mail.send` throws a checked exception, which a `Runnable` cannot, so the call goes through a generated `mail(from, recipient, subject, parts)` helper that rethrows it unchecked - carrying the original message (the outcome stamp and the row's failure line both report it) and the cause. A failed delivery therefore rolls the record back and the next tick sees an unsent row and completes it. The remaining window is a commit that fails *after* a successful send, which costs a duplicate message rather than a lost one.
- **one counter and one summary line for the combined form.** Created and mailed cannot diverge once they are one unit, and `failed` counts rows - reported in a generate summary line *and* again in a notify one, a tick with one bad row logged "failed 1" twice (the issue's secondary point).

The recipient is frozen into a `final String recipient` because the lambda may only close over an effectively final local; `to` stays the variable the row's failure line names.

A **notify-only** or **generate-only** schedule generates byte-identically to before - the reordering only moves blocks past each other when both halves are declared.

No authoring change: no new or changed intent key, so nothing on the doc sites describes behaviour this makes wrong.

## Verification

- `IntentEngineIT` - 81 tests green. Its dunning case now asserts the order the claim rests on (guard -> recipient -> `UnitOfWork.run` -> `mail(...)` -> lambda close -> `created++`), that a failed delivery leaves the lambda unchecked, and that the tick logs one summary line and no separate `sent++`.
- `IntentEmissionCoverageIT` - green. It publishes the fixture and runs the client-Java `javac` batch over the generated `BillDunningJob`, which is what proves the lambda capture, the helper and the imports actually compile; it carries the same ordering assertions.
- `ModelGenerationIT` - green (every template renders, twice).
- `engine-intent` + `ide-template` unit suites - 1282 tests green.
- `mvn formatter:format` on the changed module, then `formatter:validate` over the reactor with every `formatter-maven-cache.properties` deleted - green.

Not verified: no live SMTP failure was exercised against a running instance; the rollback-on-failed-send claim is verified by the generated code's structure, not by an end-to-end failed delivery.

Fixes #7365


delchev (commented):
The mechanics are right: the recipient is resolved before anything is written (so no history row for a message that was never addressed), the send is the unit's last act, the lambda capture is handled with `final String recipient`, `Mail.send`'s checked exception is rethrown unchecked with its message preserved for the stamp, and the double `failed` summary line is gone. The notify-only / generate-only forms are untouched. I am fairly confident this is correct for the shape that motivated the issue - a dunning reminder, where the record **is** the message.

Two things I would want settled before it lands, because the change is not scoped to that shape.

**1. A failed delivery now destroys a generated business document, and a period-keyed one may never come back.**

`generate` in a combined schedule is not always a reminder row. "Generate this month's invoice from the standing template and mail it to the customer" is the same construct, and after this change an SMTP outage rolls the **invoice** back. That is not obviously better than the bug: a document the business expects to exist is now absent because a courtesy mail could not be delivered.

It gets sharper with `run:` (#7106). A monthly tick whose send fails on the last day of the month rolls back the generation; the next tick is in the **next** month, its `run: month` range covers the new month, and the previous month's document is never generated by anyone. The old behaviour lost a mail; this loses the document *and* the mail.

The issue offered two directions and this PR took the second. The first one does not have this property: keep the write, record the delivery on the **generated** record, and have the guard match only records that were actually sent - an unsent one is then completed by the next tick rather than skipped, and nothing is ever rolled back. It costs a column on the history entity, which a reminder row wants anyway.

If the rollback shape is the deliberate choice, I would at least expect it to be conditional on what the `generate` is (a reminder vs a document), or stated in the DSL so an author of a recurring-invoice schedule is not surprised by it.

**2. SMTP I/O now runs inside an open database transaction.**

`mail(...)` is called inside the `UnitOfWork.run` lambda, so the connection and transaction are held for the whole delivery - a `connect` to an unreachable MX can sit there for the socket timeout, per row, serially through the tick. Two consequences worth a sentence in the code:

- the pool connection is occupied for the duration, and a pool or database statement/transaction timeout can now abort the write *after* a successful send - which is the "duplicate message" window the comment acknowledges, but the change makes it much wider than a normal commit;
- a slow relay turns a tick over N rows into N x (SMTP latency) with a transaction open throughout.

Neither is fatal, and both are arguably better than losing mail - but the trade is worth naming where the send sits, not only in the PR body.

Nothing wrong in the rendering itself: `subject`, `body`, `from`, `parts` are all assigned once and stay effectively final, so the capture compiles, and `IntentEmissionCoverageIT` javac-ing the generated `BillDunningJob` is the right proof for that.

**Sequencing:** #7378 edits the same no-recipient branch of this template (it adds the `skipped` outcome stamp there). The two compose semantically - after this PR the stamp lands before any write, which is if anything more correct - but they will conflict textually. Whichever goes second needs a rebase.

PR #7377, opened 2026-09-14
registry: a TypeScript transpile no longer schedules a synchronization pass (#7368)

### LINKED ISSUE #7368: registry: every TypeScript transpilation now schedules a synchronization pass - its output is written into the registry
#7303 made every create / modify / delete under `/registry/public` call `SynchronizationWatcher.force()`, which is the right generic answer for a deep *artefact* write. It also catches a writer that is not writing artefacts at all: the TypeScript transpiler.

`TypeScriptService` computes its `outDir` as the registry project directory (`IRepositoryStructure.PATH_REGISTRY_PUBLIC` + project, resolved through `repository.getInternalResourcePath`) and writes the generated `.mjs`/`.js` next to the `.ts` source. So:

- the first `/services/js/<project>/<file>.ts` request after a publish transpiles and thereby schedules a full synchronization pass;
- `TscWatcherService` re-transpiles on change, each write scheduling another.

There is no loop - a pass does not transpile - but a pass reloads every definition and checksums the whole registry, and it is now triggered by an ordinary request path that changes no artefact.

Worth deciding explicitly:

- if it is intended, nothing to do beyond a note where the watcher explains what it marks;
- if it is not, the transpiler's writes want the `RegistryMutationTracker` bracket (or an exemption by extension) so its output does not read as a registry change.

Where: `components/engine/engine-typescript/.../TypeScriptService.java`, `TscWatcherService.java`; `components/core/core-registry/.../watcher/LocalRegistryWatcher.java`.

nedelcho-delchev-tues:
## Cause

`#7303`/`#7337` made `LocalRegistryWatcher` mark the registry modified on every create/modify/delete under `/registry/public`, so any deep writer becomes self-healing. It also caught a writer that writes no artefact at all: `engine-typescript`'s esbuild (`TypeScriptService`) and `tsc --watch` (`TscWatcherService`) transpilers write their compiled output right next to the `.ts` source they compile, so the first `/services/js/<project>/<file>.ts` request after a publish, and every `tsc` re-transpile on change, scheduled a full synchronization pass - reloading every definition and checksumming the whole registry for output no synchronizer is ever keyed on (JS/TS user code loads on demand, per the synchronizer-model doc; grepping for a `FILE_EXTENSION` literal of `.js`/`.mjs`/`.ts` across every synchronizer confirms none exists).

## The change

Exempt `.js`/`.mjs` writes from `LocalRegistryWatcher`'s marking, **by extension** rather than by writer: the same is true of a hand-edited `.js` file, so the fix does not need to know which caller made the write, only what its extension is. `.ts`/`.d.ts` source files are left alone - out of this issue's scope, which is the transpiler's *output* specifically.

- `LocalRegistryWatcher.registryChanged()` now skips marking for a path whose file name ends in `.js` or `.mjs` (`hasUnwatchedExtension`), documented on the new constant and folded into the class javadoc.
- `.claude/docs/synchronizer-model.md` gets one clause added to the paragraph it already carries about this watcher, so the doc doesn't drift from the code.
- New `LocalRegistryWatcherTest` case: writing a `.mjs`/`.js` file schedules nothing, mirroring the existing "an ignored top-level folder schedules nothing" case.

## Verification

- `mvn -pl components/core/core-registry -am test` - green, 12/12 (`LocalRegistryWatcherTest` 5/5 including the new case)
- Repo-wide `mvn formatter:validate` (formatter cache wiped) - `BUILD SUCCESS`
- Full-reactor `mvn -P quick-build install -DskipTests` - green
- `RegistryDeepWriteSyncIT` (the `#7337` regression test for the general deep-write path) - still green, confirming a non-`.js`/`.mjs` deep write still schedules a pass

**Not verified:** the release-profile javadoc build - blocked by a reproducible environment issue unrelated to this change (`maven-gpg-plugin`'s bouncycastle dependency range does not resolve in this sandbox, failing before javadoc ever runs for `core-registry`). The added Javadoc follows the same paragraph/`@link` conventions already in this file, which built clean under the same profile in `#7337` days ago. Also not run: the browser-driven `TypeScriptSampleProjectsIT` (heavy, Selenide) - disproportionate for a one-extension-check change already covered at the unit level, the correct altitude for internal watcher logic.

Fixes #7368


delchev (commented):
Checked the load-bearing claim rather than taking it - the inventory of every `FILE_EXTENSION` constant across `components` is:

```
.access .bpmn .camel .confluence .csvim .extension .extensionpoint .html .java .job
.listener .markdown .md .numbers .roles .schema .scopes .table .view .websocket
```

No `.js`, no `.mjs`, so exempting them cannot make a synchronizer miss an artefact. Filtering by extension rather than by writer is the right call - the transpiler is not special, and a hand-edited `.js` is the same non-event.

Two notes, neither blocking:

- The exemption also swallows the **creation** of a directory named `*.js` (`hasUnwatchedExtension` reads the file name, not the entry type). Harmless as it stands - `registerAll` still registers it, so its children are reported on their own - but if a project ever ships a `something.js/` folder, its creation schedules nothing. A `Files.isDirectory` guard would make it exact; I would not hold the PR for it.
- `Thread.sleep(5_000)` in `aTranspiledJsFileSchedulesNothing` is the price of asserting a negative, and 5s is a reasonable floor for the macOS polling watch service. Worth keeping an eye on if more negative cases get added to this class - they do not parallelize.

`.ts` staying watched is right: a `.ts` write is authoring, and it is already covered by the publish force.

Issue #7290, opened 2026-09-10
templates: Notification.java.template stamps notify outcome only for a Mail.send failure - a relation load or attachment render failure leaves the record unstamped (#7233 sibling)


delchev:
## Describe the bug

Sibling of #7233 / PR #7278, found by grepping the other notify templates for the same shape. `Notification.java.template` (template-application-events-java) stamps `outcome:` only when `Mail.send` throws:

```java
// events/Notification.java.template
#foreach($load in $relationLoads)
        ${load.targetEntity}Entity ${load.local} = ... new ${load.targetEntity}Repository().findById(entity.${load.fkProperty});   // before the try
#end
        ...
        try {
            data = org.eclipse.dirigible.sdk.print.Print.render("${attachEntity}", language, ...);   // try/finally, no catch
        } finally { ... }
        ...
        try {
            Mail.send(from, new String[] {to}, ..., subject, parts);
            stampNotifyOutcome(entity.${notifyOutcomeKeyProperty}, null);
        } catch (Exception ex) {
            stampNotifyOutcome(entity.${notifyOutcomeKeyProperty}, ex);
            throw new RuntimeException("Failed to send notification ${name}", ex);
        }
```

A relation load that throws (a connection blip, a repository refusing), a `Print.render` failure (a broken `.print` template, a missing CMS image past the size limit is soft but a parse error is not) or the report-attach path failing propagates out of `onMessage` with **no stamp**: the broker redelivers, the `onError` log line is the only trace, and the record's `outcome:` field stays empty - the "silent by construction" state #7023 introduced `outcome:` to remove, and `-notifyFailed` never fires, so `event: { onNotifyFailed: ... }` cannot react either. PR #7278 fixed exactly this for the schedule fan-out (`Job.java.template:304-309` now stamps a load/render failure); the single-record listener did not get the same treatment. `Transition.java.template` and `Send.java.template` already open their try before the loads.

## Expected

The relation loads and the attachment render sit inside the same `try` as the send in `Notification.java.template`, so every failure of the delivery attempt stamps `failed: <reason>` and raises `-notifyFailed`, exactly as the `notify` branch of `Job.java.template` does since #7278. A test that breaks the render (an unparsable `.print`) and asserts the stamp.

Issue #7023, opened 2026-09-02
intent: a notify delivery failure is invisible - the transition succeeds, the mail never leaves, nobody is told


delchev:
### Problem

A `notify:` block is fail-soft by design (intent-assistant-guide: *"a mail problem is logged and the transition still succeeds"*). The transition part is right - the status flip is the endpoint's contract. But the only trace of the failed delivery is a server log line from `MailClient`:

```
Failed to send email from [noreply@dirigible.io] to [customer@example.com]
```

Nothing reaches the person who pressed the button, nothing is stamped on the record, and nothing downstream can react.

Concrete case (BusinessIntents sales invoices, QA pass 2026-09-02, scenario 12): Send completes, the invoice sits in SENT with SentMethod = E-mail, the customer never receives anything, and the user learns about it from the customer weeks later - the dunning schedule then mails reminders through the same broken sender, equally silently. The same shape hits every notify site: `transitions[].notify`, a process `serviceTask` notify, a `schedules[]` notify fan-out (#6356), and the report attachment mail (#6931).

### Ask

Keep fail-soft for the transition, make the outcome observable:

1. **A delivery outcome on the record** - the notify site stamps a result the way `resolves:` stamps `outcome:` (e.g. a generated `<Entity>NotificationLog` child, or `lastNotifyStatus` / `lastNotifyError` / `lastNotifyAt` columns opted in with `notify: { ..., outcome: field }`), so a list/report can show "sent / failed (reason)" per document.
2. **A signal to the acting user** - an Inbox notification or the transition's response carrying a warning ("Mail to X failed: <reason>") instead of a silent 200.
3. **Something to react to** - publish a `<Entity>-notifyFailed` topic (like `-transitioned`) so glue can set a status / open a task, and a **Resend** affordance on the record for the retry.

Fan-outs (schedules) should log per-row outcomes rather than one aggregate WARN.

Related: #6356 (notify block), #6391 (cross-model recipient), #6931 (report attachment). Suite tracker: businessintents/base-sales-invoices#19.

PR #7378, opened 2026-09-14
templates: a notify with no recipient stamps skipped, not an empty outcome (#7369)

### LINKED ISSUE #7369: templates: a notify with no recipient stamps no outcome, so it reads as never processed
#7023 added `notify: { outcome: <field> }` so a delivery is observable on the record. #7290 then closed the gap where a failure *before* the send - a relation load, a print or report render - left nothing stamped, by moving the try above those loads.

One path still stamps nothing:

```java
String to = ${toExpression};
if (to == null || to.isBlank()) {
    return;
}
```

A record whose recipient expression resolves to nothing therefore carries an empty `outcome` - indistinguishable from a record the handler never ran for. That is the state the field was introduced to remove, and it is arguably the most likely one in the field (a customer row with no e-mail address is a data problem someone must see, and a list filter on the outcome is where they would see it).

The scheduled counterpart already treats it as an outcome of its own - `Job.java.template` counts it as `skipped` and says so in the summary line - and the `transitions[]` endpoint answers `skipped` as a status. Only the record stamp has no value for it.

Suggested: stamp `skipped` (or `failed: no recipient`, if a two-value vocabulary is preferred) before the return, truncated the same way the other stamps are.

Where: `components/template/template-application-events-java/src/main/resources/META-INF/dirigible/template-application-events-java/events/Notification.java.template`.

NicoleNG18:
## Cause

`notify: { outcome: <field> }` (#7023) records a delivery on the record; #7290 closed the gap where a failure *before* the send left nothing stamped. One path still stamped nothing: when the recipient expression resolves to null/blank, the four notify templates return early **without** stamping the outcome, so a record with no recipient carries an empty `outcome` - indistinguishable from a record the handler never ran for, the exact state the field was introduced to remove. A customer row with no e-mail address is a data problem someone must see, and a list filter on the outcome is where they would see it.

## Change

- Extract a shared `writeNotifyOutcome(id, value, topic)` core and add `stampNotifySkipped(id)` - value `skipped`, **no `-notifyFailed` event** (a record with nobody to mail is a data problem to surface, not a delivery that broke) - and call it in every no-recipient branch of all four notify templates: `Notification`, `Transition`, `Send`, `Job`.
- `skipped` is the vocabulary `Job.java.template`'s summary line and the `transitions[]` endpoint already use, so the record's outcome column now reads **sent / failed / skipped** uniformly and is never empty for a processed record. The value is truncated the same way the other stamps are.

The issue names `Notification.java.template`; the same code shape (no-recipient early return without a record stamp) is in all four notify templates, so all four are fixed.

## Verification

- Extended `IntentEmissionCoverageIT`: added an `outcome:` to the step-bound notification (the `Notification` template the issue names) and assert the `skipped` stamp is emitted in the **Notification, Transition and Job** generated classes - all compiled by the publish, which is the compile proof for the refactor. The `Send` template gets the identical edit; the only `Send` call-sites in the fixture are create-triggered on the same record whose `SendBill` outcome is asserted deterministically, so coupling a `Send` outcome there would race that assertion - `Send` is covered by the identical edit rather than a separate assertion.
- `IntentEmissionCoverageIT` + `IntentEngineIT` green (82 tests, 0 failures).
- `mvn formatter:validate` BUILD SUCCESS with the formatter cache wiped.

Fixes #7369


delchev (commented):
Right fix, and extending it to all four notify templates (rather than only the `Notification` one the issue names) is the correct reading - the no-recipient early return is the same shape in each.

`skipped` with **no** `-notifyFailed` event is the right call too: the glue axis is for a delivery that broke, and a record with nobody to mail is a data problem to surface, not a failure to react to. The record's outcome column now reads sent / failed / skipped and is never empty for a processed record, which is what the issue asked for.

The `writeNotifyOutcome(id, value, topic)` extraction is a genuine de-duplication rather than a wrapper - the three call sites differ only in value and topic.

**Sequencing:** #7376 rewrites the same no-recipient branch of `Job.java.template` (it moves the recipient resolution ahead of the record write). The two compose - after #7376 the `skipped` stamp lands before anything is written, which is if anything more correct - but they overlap textually at the same few lines. Whichever merges second needs a rebase, and the rebase should keep the stamp inside the moved branch.

One question on coverage: the body says `Send` is covered "by the identical edit" rather than by an assertion, because coupling a `Send` outcome to the fixture would race the `SendBill` assertion. That reasoning is sound for the IT, but it does leave one of the four templates with no test at all. A parse-level or `ModelGenerationIT` assertion that the `Send` template *renders* the stamp would close it without touching the runtime fixture.

Issue #7343, opened 2026-09-11
data-structures: a table can have only ONE unique constraint - TableConstraint joins its owner @OneToOne, so the second one 23505s and the whole .schema fails to synchronize


delchev:
## Summary

`TableConstraint`, the `@MappedSuperclass` behind `TableConstraintUnique` / `…ForeignKey` / `…Check` / `…PrimaryKey`, joins its owning `TableConstraints` with **`@OneToOne`**:

```java
// components/data/data-structures/.../domain/TableConstraint.java
@OneToOne(fetch = FetchType.EAGER, optional = false)
@JoinColumn(name = "CONSTRAINTS_ID", nullable = false)
@OnDelete(action = OnDeleteAction.CASCADE)
@JsonIgnore
protected TableConstraints constraints;
```

The relation is genuinely **many-to-one** — one table's `TableConstraints` owns a *list* of uniques (and a list of foreign keys, and a list of checks; `TableConstraints` declares them as `List`). `@OneToOne` on the owning side makes Hibernate put a **UNIQUE index on the join column**, so `DIRIGIBLE_DATA_TABLE_UNIQUES.CONSTRAINTS_ID` can hold **at most one row per table**.

The consequence: **a table with two unique constraints cannot be synchronized.** The second insert fails 23505, `SchemasSynchronizer` fails the whole `.schema`, and the table is never created or altered.

## Reproduction

Any intent model where one entity has both a `number: { per: <Relation> }` partitioned document number (which #7017 turns into a composite `(per, number)` key) **and** an entity-level composite `unique:` (which #7092/#7116 has just made adoptable over a cross-model to-one). Concretely, `base-timesheets` on 14.54.0:

```yaml
- name: EmployeeTimesheet
  unique:
    - { fields: [ProjectTimesheet, Employee], message: "This employee already has a timesheet for this project-month" }
  fields:
    - { name: number, type: string, number: { series: Project Timesheet, per: Company, stampOn: issue }, ... }
```

The generated `timesheets.schema` is correct — two entries under `constraints.uniqueIndexes`:

```json
{ "name": "EmployeeTimesheet_ProjectTimesheet_Employee",
  "columns": ["EMPLOYEE_TIMESHEET_PROJECT_TIMESHEET", "EMPLOYEE_TIMESHEET_EMPLOYEE"] },
{ "name": "EmployeeTimesheet_Company_Number",
[...]

PR #7379, opened 2026-09-14
core-liquibase: schema-qualify the legacy CONSTRAINTS_ID unique-index sweep and drop every match (#7370)

### LINKED ISSUE #7370: core-liquibase: the legacy unique-index sweeps are not schema-qualified
#7343 added two sweeps that drop the legacy, Hibernate-named unique index on `CONSTRAINTS_ID` for deployments bootstrapped by hbm2ddl before the changelog existed. Both find the constraint through `information_schema` without naming a schema, and then drop it without qualifying the table.

PostgreSQL:

```sql
FOR r IN SELECT tc.table_name, tc.constraint_name
           FROM information_schema.table_constraints tc
           JOIN information_schema.key_column_usage k ON ...
          WHERE tc.constraint_type = 'UNIQUE'
            AND upper(tc.table_name) IN ('DIRIGIBLE_DATA_TABLE_UNIQUES', ...)
            AND upper(k.column_name) = 'CONSTRAINTS_ID'
LOOP
  EXECUTE format('ALTER TABLE %I DROP CONSTRAINT %I', r.table_name, r.constraint_name);
END LOOP;
```

`information_schema` exposes every schema the connecting role can see, while the `ALTER TABLE` is resolved through `search_path`. On a database that holds those tables in more than one schema the loop can therefore address a row found in one schema against the table in another - dropping the wrong constraint, or failing the changeset. The H2 sweep has the same gap.

Low likelihood as things stand (these are SystemDB tables), but the fix is cheap and makes the statement say what it means: add `AND tc.table_schema = current_schema()` (PostgreSQL) and the `INFORMATION_SCHEMA` equivalent for H2, and qualify the `ALTER TABLE` with the schema the row came from.

Second, smaller point in the same H2 changeset: each statement is `EXECUTE IMMEDIATE (SELECT COALESCE(MAX(...)))`, so it drops at most **one** constraint per table. If a deployment somehow carries two, one survives and nothing says so.

Where: `components/core/core-liquibase/src/main/resources/db/changelog/dirigible-system.json`, changesets `drop-legacy-named-UK_DIRIGIBLE_DATA_TABLE_CONSTRAINTS_ID-h2` / `-postgresql`.

NicoleNG18:
## Cause

The two legacy-index sweeps #7343 shipped in 14.56.0 (`drop-legacy-named-UK_DIRIGIBLE_DATA_TABLE_CONSTRAINTS_ID-h2` / `-postgresql` in `dirigible-system.json`) find the constraint through `information_schema` **without a schema filter** and then `ALTER TABLE` **without qualifying the schema**. `information_schema` exposes every schema the role can see while the `ALTER TABLE` resolves through `search_path`, so on a database that holds those tables in more than one schema the loop can address a row found in one schema against the table in another - dropping the wrong constraint or failing the changeset. The H2 sweep additionally collapses its match to `MAX(...)`, dropping at most **one** constraint per table and silently leaving a second.

## Change

Those sweeps are released, so their checksums are recorded on every deployed instance and editing them in place would fail the next upgrade's Liquibase validation - the correction is a **new** changeset (`drop-legacy-UK_DIRIGIBLE_DATA_TABLE_CONSTRAINTS_ID-schema-qualified`).

It is a Liquibase `CustomTaskChange` (`DropLegacyConstraintsIdUniqueIndexChange`) because the operation is **not portably expressible in changelog SQL**: PostgreSQL needs a `DO` loop to drop more than one, and H2 has no procedural SQL at all (`EXECUTE IMMEDIATE` runs a single statement, `ALTER TABLE` takes a single `DROP CONSTRAINT`) - so neither dialect can both schema-qualify *and* drop every match. The JDBC change:

- resolves the connection's **current schema** (`Connection.getSchema()`);
- finds every `UNIQUE` constraint on `CONSTRAINTS_ID` in the three constraint tables **within that schema** (joined within the same schema so a same-named constraint elsewhere cannot match);
- drops each with a **schema-qualified** `ALTER TABLE "schema"."table" DROP CONSTRAINT "name"`.

A schema with none yields no rows, so it is a no-op and the changelog stays idempotent. It is gated to H2 and PostgreSQL by its preconditions (`onFail: MARK_RAN`), with MSSQL out of scope exactly as the original sweeps.

## Verification

- `SystemChangelogIdempotencyTest` gains a case that plants **two** Hibernate-named unique indexes on the current schema's table **plus one of the same shape in another schema**, re-arms the corrective changeset, and asserts both current-schema indexes are dropped (**drop-all**) while the other schema's survives (**schema-qualified**). The test applies the real `dirigible-system.json` through the real `LiquibaseSystemConfig`, so it also proves the changeset parses, the customChange loads by class name, and a fresh boot runs it as a no-op.
- core-liquibase unit suite green (2/2); `mvn formatter:validate` BUILD SUCCESS with the cache wiped; the release-profile javadoc build of the module BUILD SUCCESS.
- **Not covered by a separate local test:** PostgreSQL - it uses the same portable JDBC path and is exercised by the `postgresql` CI boot leg (the no-op path).

Fixes #7370


delchev (commented):
The `CustomTaskChange` is the right answer, and the reason given for it is the real one: neither dialect can both schema-qualify and drop every match in changelog SQL. Adding a new changeset rather than editing the released ones is correct - their checksums are recorded on every 14.56.0 instance.

Details I checked and agree with: the identifier quoting doubles an embedded quote; `UPPER(...)` on both sides covers H2's upper-case and PostgreSQL's lower-case identifiers while the schema is matched as stored (`connection.getSchema()` returns `PUBLIC` on H2 and the search-path head on PostgreSQL, both of which are what `table_schema` holds); the `getSchema()` -> `database.getDefaultSchemaName()` fallback covers a driver that returns null; and a schema with no match yields no rows, so it stays idempotent. The test planting two indexes in the current schema plus one in another, and asserting drop-all *and* schema-qualification in one case, is exactly the right shape.

One gap the new changeset does not close, and I think it should be considered:

**An instance that has not upgraded yet still runs the broken sweeps first.** The 14.56.0 changesets are only recorded as run on instances that already took 14.56.0. Anything upgrading from earlier straight to a build carrying both will execute `drop-legacy-named-...-h2` / `-postgresql` - unqualified lookup, unqualified `ALTER TABLE`, `MAX(...)` - and only afterwards the corrected one. So on exactly the deployments where the old sweep can misfire, it still does; the new changeset cleans up what is left rather than preventing it.

Liquibase has the escape hatch for this: `validCheckSum: ANY` (or the recorded checksum value) on the two released changesets lets their bodies be neutralised - emptied to a no-op, or given a precondition that never passes - without failing validation on instances that already ran them. That would make the corrected change the only one that ever touches these constraints, on every upgrade path.

If you would rather not touch released changesets at all, that is a defensible call - but then it is worth a sentence in the new changeset's comment saying the old sweeps still run first on an upgrade from < 14.56.0 and why that is acceptable, so the next reader does not have to re-derive it.

Minor: the join is on `constraint_schema` + `constraint_name` + `table_name` but not `table_schema`. Equivalent in practice (a constraint's schema is its table's schema in both engines), and the explicit `tc.table_schema = ?` already pins the row - just noting it is the one column pair not tied down.

Issue #7275, opened 2026-09-09
engine-intent: no checks: kind forbids a write when a condition over a one-hop relation field holds (a forbidWhen, the reject-twin of requiredWhen)


NicoleNG18:
## Summary

The `checks:` vocabulary can require, compare and count, but it cannot express
"refuse this write when <condition> holds" - and no check condition can read a
value one hop away through a to-one relation. The two gaps together make a very
common rule inexpressible: *forbid creating/editing a child when its parent is in
a given status*.

Concrete real-world case (codbex `sales-invoices`, entity
`SalesInvoiceCustomerPayment` - the n:m payment allocation, a composition child of
`SalesInvoice`): a clerk should not be able to add a payment allocation to an
invoice whose `Status` is already `PAID`. There is no way to say this in the intent.

## Evidence (origin/master)

`components/engine/engine-intent/src/main/java/org/eclipse/dirigible/components/intent/parser/IntentParser.java`:

- The check kinds are `requiredWhen`, `exactlyOne`, `compare`, `itemsSumEqual`,
  `itemsMin` (`validateCheck`, lines 4790-4836). All of them *require*, *compare*
  or *count*; none *rejects a write on a condition*.
- `requiredWhen`'s condition is deliberately record-local:
  `validateRequiredWhenTerm` (4690-4696) rejects any `when:` term whose property is
  not "a field or to-one relation of [entity] - the condition is read off the
  record itself". So a child cannot test `SalesInvoice.Status`.
- Yet the resolver for a one-hop relation field already exists and is already used
  by the SAME check: `requiredWhen`'s `field:` is resolved through
  `ResolvePathSupport.walker(...).resolve(check.getField())`
  (`validateRequiredWhen`, 4646-4656), cross-model included. Only the *condition*
  side is missing it.

## Why the existing knobs don't cover it

- `immutableWhen` blocks *editing* an existing record and reads the record's own
  status; it does not block *creating* a new child, and a payment allocation has no
  status of its own.
- `locksWithMaster` is all-or-nothing. `SalesInvoiceCustomerPayment` sets
  `locksWithMaster: false` on purpose so payments stay addable while the invoice is
  ISSUED/SENT/PARTIAL - exactly the states where allocations happen - so turning it
  back on would over-block.
- The `invoicePaid` rollup's `capacity: payable` guard already refuses an
  *over-allocation* (a PAID invoice has zero remaining capacity, so a positive
[...]

Issue #7094, opened 2026-09-07
intent: checks: has no conditional-presence kind - 'Customer.email must be present when SentMethod is E-mail' cannot be declared, so an E-mail invoice is SENT to nobody


delchev:
### Observed

On sta (14.46.0) a sales invoice with Sent Method = E-mail whose customer has no e-mail address went through Send: status SENT, the `mailInvoice` notify step is a logged no-op without a recipient, nothing was mailed, nothing told the clerk (base-sales-invoices #24, 2026-09-06 end-user pass).

The module wanted to declare the rule where every other document rule lives:

```yaml
checks:
  - { kind: requiredWhen, field: Customer.email, when: "SentMethod == 1", status: 4, message: "Sent Method is E-mail but the customer has no e-mail address" }
```

`checks:` knows three kinds (`IntentParser.validateChecks`: "expected exactlyOne, itemsSumEqual or itemsMin"). None expresses "a field (own, or one hop over a to-one, cross-model included) must be non-empty when another field holds a value", gated at the status the Send task writes - which is exactly the shape #7065 made reach the person who pressed the button.

### Why the module cannot work around it well

- A delegate AFTER the Send task keeps its async boundary (#7065 stops the synchronous walk at an authored service task), so its refusal dead-letters instead of reaching the clerk.
- The module therefore moved the gate BEFORE Send: `checkRecipient` delegate -> decision -> `recipientHold` user task (SentMethod editable, action retry). It works, but it is ~15 intent lines + a Java class + a form + a bg catalog entry for one sentence of rule, and the clerk lands on a hold task instead of reading a refusal on the button they pressed.

### Expected

A row-level `checks:` kind with a condition and a status gate, e.g. `{ kind: requiredWhen, field: <own field | Relation.field>, when: "<Calc expression over the row>", status: <gate id>, message }`, enforced in the repository like `itemsMin` (so #7065's synchronous path applies) and on the REST write when no `status` is given (like `exactlyOne`).

Found adopting the 14.48.0 line, 2026-09-07 (BusinessIntents 14.48.0 train).

Issue #6957, opened 2026-08-27
Intent DSL: no way to guard a status consumer by HOW the status was reached - an automatic lookup and a manual task are indistinguishable


delchev:
## Problem — a DSL expressiveness gap, not a bug

When a status can be reached by **more than one path** — an automatic `resolves:` register lookup routes to `DRIVER_IDENTIFIED` on success, and an officer's user task sets the same status manually — there is no way to bind a consumer to only ONE of those paths. A `generates:` rule (or notification / integration / posting) bound to

```yaml
event: { onTransition: Fine, when: "Status == DRIVER_IDENTIFIED" }
```

fires on both, so "append a DRIVER_IDENTIFICATION_SUCCESS log row on automatic identification" also fires when the identification was manual — and the model has no way to say otherwise.

## Why the existing constructs cannot express it

- **`event.when` is a single comparison against a status seed** (`<Property> == <seed id or name>`, one term, no `||`, numeric after `StatusSymbolResolver` rewrites the name). The `resolves:` `outcome:` trace field — which stamps exactly the provenance needed (`found` / `notFound` / `ambiguous`) — is a **string** field and cannot be guarded on.
- **Splitting into two statuses** (`DRIVER_IDENTIFIED_AUTO` / `DRIVER_IDENTIFIED_MANUAL`) does not compose: a downstream at-most-once `generates:` (mint the Declaration once identification succeeded, however it succeeded) needs ONE converged status — its `when` guard is a single comparison — and two `mode: once` rules sharing the same target + back-reference are refused by design (#6813, `validateIdempotencyGuardOwnership`).

So the requirement "log which path identified the driver" is currently inexpressible; the closest authoring degrades to a shared status plus a log rule that admits (in its own description) that it fires on both paths.

## Possible directions (design discussion, not a prescription)

1. **String-literal equality in `event.when` on a declared string field** — `when: "resolution == 'found'"`. Smallest change; makes the `resolves:` trace field the provenance carrier it was designed to be. Needs a decision on quoting/escaping and on keeping the grammar a single comparison.
2. **An AND of two comparisons in `event.when`** (`Status == DRIVER_IDENTIFIED && resolution == 'found'`) — more general, more grammar.
[...]

Issue #7338, opened 2026-09-11
intent: no checks: kind constrains a field against a LITERAL - "days must be positive" is undeclarable, and every workaround for it is worse than the gap


delchev:
## What is undeclarable

`checks:` on 14.54.0 has five kinds: `exactlyOne`, `compare`, `requiredWhen`, `itemsSumEqual`, `itemsMin`. Four of them relate two things the model already names (two fields of a row, two item sums, an item count). **None of them relates a field to a LITERAL.**

So the single commonest validation in a business model has no declaration:

- `VacationDay.Days` must be **> 0** — a negative row silently inflates the parent entitlement's balance, because the roll-up sums the column verbatim.
- `VacationRequest.days` must be **> 0** before it may be submitted — a request that asks for zero leave is malformed, not a valid draft.
- a quantity, an hour count, a percentage must be `>= 0`; a percentage `<= 100`; a discount `< total`.

`{ kind: compare, field: due, op: ge, than: date }` (#7095) does exactly the right thing for two fields and stops one character short of the general case: `than:` is documented and validated as "the entity's own field", never a constant.

## Why the workarounds are worse than the gap

Three exist in the fleet today, all bad:

1. **A hand-edit of the generated controller's `validate()`.** This is what base-vacations shipped as an interim. It does not survive the next regeneration — silently — and every reviewer has to know that.
2. **A `calculatedActionOnCreate` / `calculatedActionOnUpdate` that throws.** base-vacations' `VacationWorkingDaysAction` does this for `days`. It works, but it turns a validation into a *calculation*, it only fires on the field that declares it, and the message reaches the caller as whatever the action's exception carries rather than as an authored refusal.
3. **Not enforcing it at all**, and documenting the hole.

## What is asked

Let a check compare a field with a constant. The smallest shape reuses `compare` and its whole implementation:

```yaml
checks:
  - { kind: compare, field: days, op: gt, value: 0, message: "A vacation request must cover at least one working day" }
  - { kind: compare, field: discountPercent, op: le, value: 100, message: "A discount cannot exceed 100%" }
```

[...]

PR #7414, opened 2026-09-16
docs(intent): correct the guard grammar and document the conditional checks

TIVMOF:
## The defect

#6996 (mine) added this sentence to the `resolves:` section:

> A `generates:` / `postings:` / process `trigger:` guard compares a **status** and nothing else — `when:` cannot read the `outcome:` string

It is false, and it was false when it was written — #6957 had already made a `when:` an AND-ed list of equality terms over the record's own columns, status or not. Posting the counter-example to `/services/ide/intent/validate`:

```
issues: []
parsed event when: ["Status == 2", "resolution == found"]
```

A status term **and** a string term on an ordinary field, accepted.

## Why it mattered

The sentence was not inert. An intent generated from a full process description split one lookup's two failure outcomes into two statuses because the guide said it had to — and therefore needed two otherwise identical fallback processes and two otherwise identical audit rules. It used `resolution == found` for its success guard a few lines below, so it had the capability in hand and still obeyed the prohibition.

Routing both failures to one status and separating the audit rows on the outcome removes a process, a create-from and a seeded status, and validates clean:

| | before | after |
|---|---|---|
| processes | 2 | **1** |
| generates | 7 | **6** |
| fine statuses | 4 | **3** |

## What changed

**The rule is restated as the choice it is** — one status plus an outcome term when the two failures are handled the same way and differ only in what the trail records; separate statuses when they need different flows, since a process binds at most one trigger and a guard list is AND-ed, never OR-ed. One status *and no outcome term* remains the thing that is always wrong.

**The grammar is now written down.** The reason a wrong sentence about it was believable is that it was documented nowhere — the axis section showed `when: "Status == 2"` and left the rest to inference. Now stated: the operators; that the property is the record's own and why a dotted path is refused; that a status name is already its seed id; that only string/integer/boolean are guardable and why a decimal is refused rather than compared (a boxed cross-type comparison is silently always-false); that a list is AND-ed with no OR; and that `onTransition` requires a guard while the other axes do not.

**`requiredWhen` (#7094) and `forbidWhen` (#7275) get real entries.** They appeared only as a subordinate clause inside the `compare` bullet — the parser accepts seven check kinds and the guide taught four. Both now document the point that decides whether they work at all: **without the `status:` gate a check runs in the controllers only**, so a rule meant to gate a status a workflow sets needs the gate, because `setField`/`setRelationField` are repository writes that never reach a controller.

That is the check that expresses *"this may be empty in a draft but not once it reaches STATUS"* — which `required: true` cannot say, and which any app that fills a relation through a user task needs to stop a task completion marking a record done with the value still empty.

Two quick-reference lines were added for the same two shapes.

## Verification

Every YAML added here was validated against a running instance rather than read off the source:

- the AND-ed outcome guard — `issues: []`, parsed as `["Status == 3", "resolution == notFound"]`
- the `requiredWhen` gate — `('requiredWhen', 'payer', 'Status == 2', 2)`, both symbols resolved to the seed id
- the `forbidWhen` one-hop parent condition — `('forbidWhen', None, 'Invoice.Status == 2', None)`

Documentation only; no code changes.


delchev (changes requested):
Reviewed against `origin/master` at 6bbb794de7. The `resolves:` correction is right and the `requiredWhen` / `forbidWhen` entries are accurate, but the new **"`when:` - the guard grammar"** section makes two claims the parser contradicts, one of them directly against the paragraph immediately above it in the same guide section.

### 1. "An `onTransition` binding **requires** a guard; on the other axes it is optional" is false

On `notifications`, `integrations` and `outbound` the validator returns as soon as `when` is absent, whatever the trigger kind (`validateEventGuard`, `IntentParser.java:3099`). The guide says so itself five lines earlier, with an `onTransition` example commented `# the guard is optional here`. The guard is mandatory for `onTransition` only on `generates` (`validateGeneratesWhen`, `requireStatusGuard`) and `postings` (`IntentParser.java:7724`). An assistant reading the more specific later sentence will add guards nobody asked for, or refuse "notify on any status change".

### 2. The section describes one grammar for the whole axis, but the constructs it generalizes over each have their own

The typed grammar it documents (`==|!=`, string/integer/boolean, ANDed list, wrong type refused) is exactly what `validateGuardTerm` + `CheckSupport` enforce - and only the three glue consumers use it. The others differ:

- **`postings.event.when`**: scalar only, `<Property> == <number>`. A list is refused with *"event when does not take a list here"* (`IntentParser.java:7718`), and there is no string term. So *"nothing restricts a guard to the status column"* is false for postings, and the sentence this PR deletes was **correct** for them (the PR body names postings among the axes the old sentence misdescribed).
- **`generates.event.when`**: a list, but at most **one** numeric term, `==` only (`WHEN_STATUS_TERM`), plus `<StringField> ==|!= <literal>` terms on **string/text** fields only - an integer or boolean field is refused: *"only the source's string/text fields can carry a literal guard"* (`IntentParser.java:8664`); a duplicate property is refused. So both `==|!=` and "string, integer and boolean are guardable" overstate here.
- **process `trigger.when`** and a `wait` step's `when`: no parser validation at all; both render through the untyped `NotificationSupport.guard(Object)` (`GlueIntentGenerator.java:249`, `:5274`), which turns an unparseable term into `true`. The "refused rather than compared" guarantee does not hold there.
- `resolves.event.when` and `transitions[].when` are scalar only.

The section sits under *the event axis*, which the guide defines as the three glue consumers plus event-driven `generates`. Scoping it is enough: the typed grammar for the three; `generates` takes exactly one `<Status> == <name|id>` term plus string-field terms; `postings` takes exactly one status term and no list. The rewritten `resolves` example (a `generates` with one status term and one term on the string field `resolution`) is valid and renders as `Objects.equals(source.Resolution, "notFound")`.

### 3. Minor

- *"`message` is mandatory on both"* holds only for `forbidWhen` (`IntentParser.java:5293`); `validateRequiredWhen` never checks it.
- Worth one clause: a `generates` guard on an **editable** string field gets a Generate warning, and `readOnly: true` on the `outcome:` field is the fix (`warnIfGuardFieldIsEditable`).

### Confirmed correct

`compare` does take an optional `status:` gate on master since #7338, so "the same optional gate as `compare`" is right. The controller-vs-repository routing by gate presence matches `ModelParameterProcessor`. `forbidWhen` refuses a `field`, allows one hop, and requires the EntityStatus relation when gated. A bare to-one as a `requiredWhen` `field` resolves to the FK read, so `field: driver` works.

delchev:
Pushed b300d614bf onto this branch with the corrections from the review: the grammar section is now scoped per construct (glue consumers: typed `==|!=` list, guard optional on every trigger kind; `generates`: one `==` status term plus string-field terms; `postings`: the status term alone, no list; `trigger:` / `wait`: unvalidated at parse), the `onTransition`-requires-a-guard sentence is gone, the `resolves` paragraph says the outcome-reading guard is a `generates` one, and `message` is stated mandatory only on `forbidWhen`. Every statement was re-checked against the parser on master (`validateEventGuard`, `validateGeneratesWhen`, the postings block, `TriggerSupport.triggerWhen`).

delchev (approved):
Corrections pushed in b300d614bf; the guide now matches the parser.

PR #7438, opened 2026-09-18
csvim: the IDENTITY restart after a seed import quotes the schema instead of allow-listing it (#7435)

### LINKED ISSUE #7435: csvim: the IDENTITY restart after a seed import is skipped for every tenant schema - SAFE_IDENTIFIER rejects the schema name the platform itself generates
## What happens

A CSV seed import always writes explicit primary keys, which leaves the table's IDENTITY counter at
its initial value. `CsvimProcessor` knows this and repairs it right after the load:

```java
// CsvimProcessor.java:240-247
updateSequence(csvFile, connection, countAll);
if (countAll > 0) {
    // CSV inserts always carry an explicit PK value; the DB's IDENTITY counter is therefore
    // not advanced by the load. Bump it to MAX(col)+1 so a subsequent INSERT … DEFAULT
    // (e.g. from a Hibernate @GeneratedValue(IDENTITY) entity) doesn't collide on the seeded
    // rows.
    restartIdentityColumns(dataSource, connection, targetSchema, tableName);
}
```

But `restartIdentityColumns` refuses to run unless the schema and table names match an allow-list:

```java
// CsvimProcessor.java:304-311
private void restartIdentityColumns(DirigibleDataSource dataSource, Connection connection, String schema, String tableName) {
    if (!SAFE_IDENTIFIER.matcher(String.valueOf(schema)).matches()
            || !SAFE_IDENTIFIER.matcher(String.valueOf(tableName)).matches()) {
        logger.warn("Skipping IDENTITY restart — schema/table name not a safe SQL identifier: [{}].[{}]", …);
        return;
    }

// CsvimProcessor.java:471
private static final Pattern SAFE_IDENTIFIER = Pattern.compile("[A-Za-z_][A-Za-z0-9_]*");
```

**A tenant schema can never match that pattern.** The platform assigns a tenant a random UUID as its
id (`TenantEndpoint.java:108` — `tenant.setId(UUID.randomUUID().toString())`) and derives the schema
name from it (`DefaultDataSourceProvisioning.java:166-169` — `return tenant.getId().toUpperCase()`).
A UUID fails on the hyphens, and two times out of three on the leading digit as well. So for every
tenant the platform provisions itself, the counter is left where the load found it.

Observed on PostgreSQL, with the schema and table names replaced by neutral equivalents — the only
difference between the two lines is a hyphen in the schema name:

```
[WARN] [tenant-initialization] CsvimProcessor - Skipping IDENTITY restart — schema/table name not a safe SQL identifier: [my-tenant].[ORDERS]
[INFO] [tenant-initialization] CsvimProcessor - Advanced IDENTITY counter on [mytenant.ORDERS.ORDER_ID] to [7]
```

## The consequence

The first row a user creates through a generated UI or API in that tenant asks Hibernate's
`identity` generator for a value, PostgreSQL hands back `1`, and the insert dies on a seeded row:

```
ERROR: duplicate key value violates unique constraint "ORDERS_pkey"
  Detail: Key ("ORDER_ID")=(1) already exists.
```

What makes this expensive to diagnose is that it *heals itself*. A failed insert still consumes a
value, so each retry moves one id further — 1, 2, 3 … — until the counter clears `MAX(seeded)` and a
create finally succeeds. With six seeded rows the user sees six failures and then normal behaviour,
which reads as flakiness rather than as a deterministic defect. The only durable signal is the WARN,
and nothing surfaces it to the person doing the creating.

This is the tenant-schema remainder of #3482, which reported the same collision for the default
schema. `restartIdentityColumns` closes that case; it does not reach this one.

## Why the default tenant is fine, and why CI stays green

`PUBLIC` matches `SAFE_IDENTIFIER`, so in the default tenant the restart runs and everything works.
The regression test is pinned to exactly that case — `CsvimIdentityRestartIT.java:87` declares
`"schema": "PUBLIC"` and asserts the next generated id is 6 — and there is no multitenant variant
anywhere in `tests/`. The one path that works is the only one covered.

## Two sibling guards fail the same way

`csvSuppliesIdentityColumn` (`:416-423`) and `setMssqlIdentityInsert` (`:448-455`) carry the
identical `SAFE_IDENTIFIER` test and are therefore also dead on a tenant schema. On MSSQL that is
worse than a stale counter: without `SET IDENTITY_INSERT … ON` the engine rejects an explicit value
for an identity column outright, so the seed import itself fails rather than merely leaving the
counter behind.

## Suggested fix

Quote the identifiers instead of allow-listing them. The allow-list was written for a world of
unquoted identifiers and was never reconciled with the platform's own schema naming, while the sinks
already quote:

- **PostgreSQL** (`:390-395`) emits `setval(pg_get_serial_sequence('"<schema>"."<table>"', '<column>'), n, false)`
  — already inner-quoted, and would work unchanged for a UUID schema.
- **H2** (`:386-389`) already builds its DDL with `connection.getMetaData().getIdentifierQuoteString()`.
- **MySQL/MariaDB** (`:399-400`) already uses backticks.
- **MSSQL** (`:396-398`) passes `'<schema>.<table>'` to `DBCC CHECKIDENT` unquoted, so that one needs
  `[schema].[table]`.

The residual injection risk is a quote character *inside* a name, which doubling handles. If the
allow-list is kept instead, it must at minimum accept what the platform's own provisioning
generates.

Whatever the mechanism, a skipped restart should not be a WARN that only a log reader ever sees when
the tenant it applies to is about to serve traffic.

## Acceptance criteria

- After a seed import into a non-default tenant schema, an INSERT that omits the identity column
  receives `MAX(seeded)+1`, not `1`.
- A multitenant variant of `CsvimIdentityRestartIT` covers it — the current test cannot fail on this.
- Identifiers reach the DDL correctly quoted per dialect, and no identifier is concatenated into SQL
  unquoted; injection safety is preserved without rejecting legitimate names.
- `csvSuppliesIdentityColumn` and `setMssqlIdentityInsert` work on the same names, so an MSSQL seed
  import into a tenant schema succeeds.
- A skip remains possible only for a dialect with no supported restart form, and says so.

## Evidence

| claim | file:line |
|---|---|
| the restart is called after every non-empty load | `components/data/data-csvim/src/main/java/org/eclipse/dirigible/components/data/csvim/processor/CsvimProcessor.java:240-247` |
| …and bails out on the allow-list | same file `:304-311` |
| the allow-list | same file `:471` |
| the four dialect branches | same file `:375-409` |
| the two sibling guards | same file `:416-423`, `:448-455` |
| CSV inserts carry explicit PKs | `components/data/data-csvim/src/main/java/org/eclipse/dirigible/components/data/csvim/processor/CsvProcessor.java:95-113` |
| a tenant id is a random UUID | `components/core/core-tenants/src/main/java/org/eclipse/dirigible/components/tenants/endpoint/TenantEndpoint.java:108` |
| the tenant schema name is that id, uppercased | `components/data/data-sources/src/main/java/org/eclipse/dirigible/components/data/sources/provisioning/DefaultDataSourceProvisioning.java:166-169` |
| a modelled `generated: true` PK becomes a native identity column on PostgreSQL | `components/data/data-structures/src/main/java/org/eclipse/dirigible/components/data/structures/synchronizer/table/TableCreateProcessor.java:88`, `modules/database/database-sql-postgres/src/main/java/org/eclipse/dirigible/database/sql/dialects/postgres/PostgresSqlDialect.java:115-117` |
| the generated entity asks the DB for it | `components/data/data-store-java/src/main/java/org/eclipse/dirigible/components/data/store/java/hbm/JavaEntityToHbmMapper.java:183-189` |
| the test pins the default schema | `tests/tests-integrations/src/main/java/org/eclipse/dirigible/integration/tests/api/CsvimIdentityRestartIT.java:87` |

iliyan-velichkov:
## The cause

A CSV seed import always writes explicit primary keys, so the table's IDENTITY counter is left where the load found it; `CsvimProcessor` repairs it right after with `restartIdentityColumns`. That repair - and the two MSSQL `IDENTITY_INSERT` guards beside it, `csvSuppliesIdentityColumn` and `setMssqlIdentityInsert` - were gated on `SAFE_IDENTIFIER`, an allow-list of bare unquoted identifiers (`[A-Za-z_][A-Za-z0-9_]*`).

**A tenant schema can never match it.** The platform assigns a tenant a random UUID (`TenantEndpoint`) and derives the schema name from it (`DefaultDataSourceProvisioning`), so the name carries hyphens and, two times out of three, a leading digit. For every tenant the platform provisions the restart was skipped with only a WARN, and the first records a user created through a generated UI or API collided with the seeded rows on PK = 1 - one failure per seeded row, healing itself afterwards, which reads as flakiness rather than as a deterministic defect. `PUBLIC` matches the allow-list, which is why the default tenant was fine and why the existing regression test - pinned to `"schema": "PUBLIC"` - could not fail on this.

On MSSQL the same gate is worse than a stale counter: without `SET IDENTITY_INSERT ... ON` the engine rejects the explicit ids the seed import carries, so the import itself fails.

## The change

The allow-list is replaced by quoting, which is what most of the sinks already did. Each identifier is emitted in the form its dialect defines, with the closing delimiter doubled, and any name that reaches a SQL string literal single-quote-escaped:

| sink | form |
|---|---|
| `SELECT MAX(...)`, H2 `ALTER TABLE ... RESTART WITH` | the driver-reported quote string, SQL-standard `"` when a driver reports none |
| PostgreSQL `pg_get_serial_sequence` | the inner-quoted `"<schema>"."<table>"` inside its string literal (unchanged in shape - it already worked for a UUID schema) |
| MSSQL `DBCC CHECKIDENT`, `SET IDENTITY_INSERT` | `[schema].[table]` - `DBCC CHECKIDENT` was the one sink passing an unquoted name |
| MySQL / MariaDB | backticks |

Three smaller consequences:

- **A skip is now possible only for a dialect with no supported restart form**, and says so - at WARN, naming what is left behind, rather than the INFO line it was.
- **Absent a schema the table is emitted bare**, resolved by the connection's own schema, instead of the literal `"null"` the old `String.valueOf(schema)` produced and then failed on.
- The catch blocks in the touched methods now log the throwable, per the repo's logging convention.

## Verification

- **`CsvimIdentityRestartTenantIT` (new)** - provisions a tenant, seeds five rows into its schema through the ordinary `.table` / `.csv` / `.csvim` artefacts, and asserts an INSERT that omits the identity column receives 6. It first asserts the schema really is not a bare identifier, so it stays a test about this defect. Run against the unfixed processor it fails with exactly the reported sequence: `Skipping IDENTITY restart - schema/table name not a safe SQL identifier: [47EC376E-DFB4-...]` followed by a PK violation on id 1.
- Green on **H2** and on **PostgreSQL 16** (local container, CI's datasource variables): `CsvimIdentityRestartTenantIT`, `CsvimIdentityRestartIT`, `CsvimReimportIT`; `CsvimIT` green on H2. The PostgreSQL run shows `Advanced IDENTITY counter on [D565DB1E-F293-40F7-9B52-1D212B50DAF7.CSVIM_IDENT_TENANT_BOOK.BOOK_ID] to [6]`.
- **`CsvimProcessorIdentitySqlTest` (new, unit)** pins the emitted statement for H2, PostgreSQL, MSSQL, MySQL and MariaDB, the unsupported-dialect `null`, the bare-table form and delimiter doubling in all three quoting styles. MSSQL and MySQL/MariaDB have no CI leg, so this is the only place their statements are checked at all.
- `mvn formatter:validate` with the formatter cache wiped, and the `release`-profile javadoc build on `components/data/data-csvim -am`.

**Not verified:** no MSSQL or MySQL server was run - those two dialects are covered by the unit test's SQL text only. `IDENTITY` DDL emission is suppressed on MSSQL by `TableCreateProcessor`, so the integration tests skip there as the existing one already did.

Fixes #7435


iliyan-velichkov:
## CodeQL findings on this PR

Six alerts, all on `CsvimProcessor.java`. None of them is new behaviour: they are the **same sinks that were already dismissed on master** (`java/sql-injection` 10662 / 10663 / 10670 / 10672 / 10711, `java/log-injection` 10712-10724, all "won't fix" with no comment). The `SAFE_IDENTIFIER` allow-list this PR removes was never a barrier for either query - `QueryInjectionFlowConfig` models no sanitizer but `SimpleTypeSanitizer`, and `LogInjectionConfig` only models line-break removal - it merely held the sinks at different line numbers, so moving them stopped the old dismissals from matching and the findings were re-raised unchanged.

### 5 x `java/log-injection` - fixed in code (b341a10)

These were worth fixing rather than dismissing again. The values reaching `logger.*` were already routed through the class's own `sanitize()`, but it stripped `[CR LF TAB]` as a hand-rolled character class, which

- **misses the Unicode line terminators** NEL (U+0085), LS (U+2028) and PS (U+2029), each of which splits a log line just as well as LF, and
- is not a shape CodeQL models, so it read as no sanitizer at all.

`sanitize()` now does `replaceAll("\\R", "_")` - the regex Java defines as "any line terminator", and the one `LineBreaksLogInjectionSanitizer` accepts - with tabs keeping their own pass. So the five findings go away because the value really is sanitized, not because someone asserted it was. `CsvimProcessorIdentitySqlTest.aLoggedIdentifierCarriesNoLineTerminatorAndNoTab` pins the property, NEL / LS / PS included.

### 1 x `java/sql-injection` (11431) - dismissed as a false positive

No code change can clear this one: the query's configuration models no barrier other than `SimpleTypeSanitizer`, so no string-level mitigation is visible to it.

Why it is not a vulnerability:

- **A schema, table or column is an identifier, not a value.** It cannot be parameter-bound, so quoting is the only mitigation the position admits - which is exactly what this PR introduces.
- **Every sink quotes**, in the form its dialect defines: the driver-reported delimiter for the `MAX()` probe and the H2 `ALTER`, the inner-quoted name inside PostgreSQL's `pg_get_serial_sequence` literal, `[x]` for `DBCC CHECKIDENT` and `SET IDENTITY_INSERT`, backticks for MySQL/MariaDB. The **closing delimiter is doubled** (`CsvimProcessor.quote`) and **single quotes are doubled** wherever a name reaches a SQL string literal (`CsvimProcessor.literal`), so no name can terminate its own quoting. `CsvimProcessorIdentitySqlTest.anEmbeddedDelimiterIsDoubledRatherThanClosingTheQuote` pins that in all three quoting styles.
- The PR **strictly improves** the situation the alert is about: before it, `DBCC CHECKIDENT` was the one sink concatenating a bare unquoted name into its argument.
- The value is **not remote input**. It is the `schema` of a `.csvim` artefact authored by a DEVELOPER and published into the registry, or `connection.getSchema()` read back from the driver. A principal who can publish a `.csvim` already reaches the database through the SQL console and the JS/Java engines, so this sink adds no reachable surface.

PR #7441, opened 2026-09-18
#7415 - Support for front-ends hosted outside Dirigible

### LINKED ISSUE #7415: Support frontends hosted outside Dirigible
## Motivation

Some applications built on Dirigible need a frontend that is not served by the platform itself:

- a single page application deployed on static hosting under another origin
- a native mobile or desktop app
- a mobile or desktop app that packages the same web frontend and displays it in a webview, whose origin (for example `tauri://localhost`, `http://tauri.localhost`, `capacitor://localhost`) is not same site with the backend.

All three cases call the platform's REST endpoints (`/services/java/**`, `/public/java/**`, the generated entity and report APIs) and the STOMP broker from a different origin. The second case cannot rely on the session cookie at all, because the platform's cookie becomes a third party cookie inside the webview, which Safari, WKWebView and Android WebView block by default. Such clients therefore authenticate with bearer tokens issued by the configured identity provider.

Today neither works on the `cognito` and `keycloak` profiles. This issue asks for a small, profile agnostic feature set so that the backend can serve mobile apps and separately hosted SPAs in addition to the platform hosted web UIs. Everything proposed is opt in and changes nothing when the new configuration is absent.

## Current behaviour

1. **No CORS on the OAuth2 chains.** `CognitoSecurityConfiguration` and `KeycloakSecurityConfiguration` never call `HttpSecurity.cors(...)`. The existing `CorsConfigurationSourceProvider` (`components-core-base`) is wired only into `BasicSecurityConfig` (and the snowflake chain), it is hardcoded to `allowedOriginPatterns("*")` with `allowCredentials(true)`, its allowed methods list omits `OPTIONS`, and there is no `DIRIGIBLE_CORS*` configuration anywhere. A cross origin request to `/services/**` on an OAuth2 profile fails in the browser before any handler runs.

2. **Preflights have no handler.** `JavaEndpoint` maps GET, POST, PUT, PATCH and DELETE only. Without a `CorsFilter` in the chain, a preflight `OPTIONS` request falls through to the authorization rules and is answered with a redirect to the login page.

3. **Programmatic requests get a login redirect instead of a 401.** `BasicSecurityConfig` has an `isProgrammaticRequest` matcher (`Sec-Fetch-Mode`, `X-Requested-With`) wired to `HttpStatusEntryPoint(401)`. The OAuth2 chains have no such entry point, so an unauthenticated XHR or fetch receives a 302 to `DIRIGIBLE_SECURITY_LOGIN_PAGE` (or the provider's authorization endpoint) and an HTML body.

4. **Bearer tokens are accepted but produce a different identity than the cookie session.** Both OAuth2 chains configure `oauth2ResourceServer().jwt()` with `NimbusJwtDecoder.withJwkSetUri(...)` and a `JwtAuthenticationConverter` whose granted authorities converter is `ScopeRoleJwtAuthoritiesConverter` (`scope` and `scp` claims mapped through the scope to role mappings). Consequences:
   - the principal name is the default `sub` claim, while `oauth2Login` on the same chain uses `user-name-attribute=email` (cognito) or `preferred_username` (keycloak). The same person therefore has a different `User.getName()` depending on whether they arrive with a cookie or a token, which breaks any application that stores the principal name (ownership columns, audit `CreatedBy`, `convertAndSendToUser` routing),
   - the identity provider's group claim (`cognito:groups`, `groups`), which `oauth2Login` maps to roles through `TenantAwareAuthoritiesMapper`, is ignored on the bearer path, so a token caller holds no roles,
   - the decoder validates the signature and expiry only. There is no issuer, audience or (for Cognito) `token_use` validation, and the JWKS location cannot be overridden, so the path cannot be exercised against a local mock issuer in automated tests.

5. **STOMP is same origin and cookie only.** `WebsocketConfig` registers `/stomp` (raw and SockJS) without `setAllowedOriginPatterns`, so Spring's default origin check rejects cross origin handshakes, and the STOMP session's user comes exclusively from the HTTP session on the handshake. A bearer token client has no way to authenticate the socket, and `SimpMessagingTemplate.convertAndSendToUser` never reaches it.

## Proposed changes

All of them should live in shared code used by every security chain (cognito, keycloak, basic, snowflake), configured through `DirigibleConfig`, and stay inert when the configuration is absent.

### 1. Configurable CORS

- `DIRIGIBLE_CORS_ALLOWED_ORIGINS`: comma separated list of origin patterns (Spring `allowedOriginPatterns`). Empty (default) means CORS stays exactly as it is today on every chain.
- `DIRIGIBLE_CORS_ALLOW_CREDENTIALS`: default `false` (bearer clients do not need cookies, and credentials with a wildcard must stay impossible).
- `DIRIGIBLE_CORS_ALLOWED_METHODS`: default `GET,HEAD,POST,PUT,PATCH,DELETE,OPTIONS`.
- `DIRIGIBLE_CORS_ALLOWED_HEADERS`: default `Authorization,Content-Type,Accept,X-Requested-With`.
- `DIRIGIBLE_CORS_EXPOSED_HEADERS`: default `Content-Disposition`.
- `DIRIGIBLE_CORS_MAX_AGE`: default `3600`.

When the origin list is set, every chain calls `http.cors(...)` with a `CorsConfigurationSource` built from these values, registered for `/**`. Spring Security's `CorsFilter` then answers preflights before authorization, which also covers `JavaEndpoint`, `WebEndpoint` and the generated controllers without touching them. The `.access` constraint filter must not interfere with preflights (they carry no credentials by definition). Once this exists, `BasicSecurityConfig` should use the same source instead of the hardcoded provider, keeping its current permissive behaviour as the fallback when nothing is configured.

### 2. Programmatic 401 on the OAuth2 chains

Reuse the basic chain's `isProgrammaticRequest` matcher, extended with "an `Authorization: Bearer` header is present" and "`Accept` prefers `application/json`", and register `HttpStatusEntryPoint(401)` for it through `exceptionHandling().defaultAuthenticationEntryPointFor(...)`. The response should carry `WWW-Authenticate: Bearer` and a small JSON body. Browser navigations keep today's redirect to the login page, so the platform's own web UIs are unaffected.

### 3. Consistent identity on the resource server path

- **Principal claim.** `DIRIGIBLE_OAUTH2_JWT_PRINCIPAL_CLAIM`, defaulting to the chain's `user-name-attribute` (`email` on cognito, `preferred_username` on keycloak), applied with `JwtAuthenticationConverter.setPrincipalClaimName`. A cookie session and a bearer token then yield the same `User.getName()`.
- **Group based roles.** Map the provider's group claim on the bearer path through the same `TenantAwareAuthoritiesMapper` and group claim name (`cognito:groups`, `groups`) that `oauth2Login` already uses, so that roles are identical on both paths. The existing scope to role mapping stays additive.
- **Validation.** Build the decoder with issuer validation (`withIssuerLocation` or the configured issuer), an audience validator over `DIRIGIBLE_OAUTH2_JWT_AUDIENCES` (comma separated, default the configured client id) and, on the cognito chain, a `token_use` validator (`DIRIGIBLE_COGNITO_JWT_TOKEN_USE`, default `id`, since only the ID token carries the e-mail and the groups).
- **Overridable endpoints.** `DIRIGIBLE_OAUTH2_JWT_ISSUER_URI` and `DIRIGIBLE_OAUTH2_JWT_JWK_SET_URI` so integration tests can run the chain against a local mock OIDC issuer.
- Confirm that `UserFacade.getName()` and `UserFacade.isInRole()` behave correctly when the `Authentication` is a `JwtAuthenticationToken`.

### 4. STOMP for cross origin and bearer clients

- `registerStompEndpoints` applies `setAllowedOriginPatterns` from `DIRIGIBLE_CORS_ALLOWED_ORIGINS` (same list, same default of unchanged behaviour when empty).
- An inbound `ChannelInterceptor` handles the `CONNECT` frame: when an `Authorization: Bearer` native header is present, decode and convert it with the decoder and converter of change 3 and call `StompHeaderAccessor.setUser(authentication)`. Without the header, keep the handshake principal as today. A rejected token fails the CONNECT with a STOMP `ERROR` frame.

### 5. Optional: bearer to session exchange

`POST /login/token` accepting a valid bearer token and establishing a regular cookie session for the same principal and authorities. This lets a bearer client deep link into the platform hosted web UIs (the shell, the generated admin pages) without a second sign in. Nice to have, separable from the rest.

## Acceptance criteria

- With `DIRIGIBLE_CORS_ALLOWED_ORIGINS` unset, every chain behaves exactly as in 14.33.0 (CORS, entry points, STOMP, JWT handling).
- On the cognito and keycloak profiles with a listed origin: a preflight `OPTIONS` to `/services/java/...` answers 200 or 204 with the matching `Access-Control-Allow-*` headers before authorization runs, a request from an unlisted origin is rejected by the `CorsFilter`, and `Content-Disposition` is exposed.
- An unauthenticated request with `Accept: application/json` (or a bearer header) to `/services/**` answers 401 with `WWW-Authenticate: Bearer` and a JSON body. An unauthenticated browser navigation still redirects to the login page.
- A request carrying a valid ID token yields the same `User.getName()` and the same roles as a cookie session of the same user. A token with a wrong `aud`, a wrong `iss`, an unexpected `token_use` or an expired `exp` answers 401.
- A STOMP `CONNECT` over a cross origin handshake from a listed origin, carrying a valid bearer header, is accepted, and a message sent through `convertAndSendToUser(<principal name>, ...)` reaches that session. A `CONNECT` with an invalid token is rejected.
- The environment variables are documented on the Environment Variables page.

## Testing notes

The bearer path can be covered without a real identity provider by pointing `DIRIGIBLE_OAUTH2_JWT_ISSUER_URI` and `DIRIGIBLE_OAUTH2_JWT_JWK_SET_URI` at a mock OIDC issuer (for example `navikt/mock-oauth2-server`) that mints tokens with arbitrary `email` and group claims, and by running the browser side of the tests from a second origin served by a plain static web server.

## Out of scope

- Returning tokens from `POST /login/native` (clients that need tokens obtain them from the identity provider with a public app client, the platform only has to accept them).
- Cookie attribute changes (`SameSite=None`), since the bearer approach removes the need for cross site cookies.
- CSRF changes. The chains keep `csrf().disable()`, and `DIRIGIBLE_CORS_ALLOW_CREDENTIALS` defaults to `false`, so no cookie authenticated cross origin write becomes possible by default.

### LINKED ISSUE #7446: Tenant selection for bearer requests under TOKEN_GROUPS
## Context

#7415 / #7441 added bearer-token support for external frontends. Under `DIRIGIBLE_TENANT_RESOLUTION_STRATEGY=TOKEN_GROUPS`, a bearer request is always resolved to the **default tenant**, regardless of which tenants the token's groups grant access to:

- `TenantExtractor.determineSelectedTenant` returns the default tenant unconditionally whenever the request carries a bearer `Authorization` header, before any session-based selection is even looked at.
- `ResourceServerJwtSupport` additionally refuses a token whose groups carry no *global* role - a user who is a member of one or more tenants but holds no global role is rejected outright.

Together, these mean an external frontend on a `TOKEN_GROUPS` deployment can only ever act as platform staff (global roles). It cannot reach any of the tenants the calling user's groups actually assign them to.
For a deployment whose main reason to adopt `TOKEN_GROUPS` is to serve several tenants behind one host, this defeats the multi-tenant use case for exactly the client #7415 was built for.

## Why the request can't already carry the answer

Session-based tenant selection (`TenantSelectionManager.selectTenant`) works because it runs *after* authentication, validates the requested tenant against the authenticated principal's own groups, and stores the choice in the session.
A bearer request has no session to store a choice in, and the tenant is currently resolved by a filter that runs **before** authentication - so there is no principal to validate against yet at the point the decision is made.

## Wanted

A way for a bearer request to select a tenant, validated the same way a session selection is:

- A header (e.g. `X-Tenant-Id`) or a token claim naming the tenant, read once the request is authenticated - not by the pre-authentication filter that decides today.
- Validated against the token's own groups, using the same membership check `TenantSelectionManager.selectTenant` applies to a session selection (403 if the token's groups don't include that tenant, 409 if the tenant isn't `PROVISIONED`).
- Absent the header/claim, current behavior is preserved: default tenant, global roles only - so this is additive and doesn't change any deployment that isn't opting in.

## Where to start

- `TenantExtractor.determineSelectedTenant` (`core-tenants`) - the bearer branch that currently short-circuits to the default tenant.
- `TenantSelectionManager.selectTenant` (`security-oauth2`) - the validation logic to mirror.
- `ResourceServerJwtSupport` - the refusal for a token with only tenant-scoped groups, which would need to become conditional on whether a tenant was actually requested and resolved.

## Not in scope here

This does not change `SUBDOMAIN` mode, which is host-based and unaffected. It also doesn't change the "no global role -> refused" default for a bearer request that requests no tenant - only a request that explicitly asks for one it's a member of should be let through on tenant roles alone.

StanZGenchev:
Front-ends hosted outside Dirigible (SPAs, native mobile/desktop apps, Electron/Tauri shells, etc) call the REST API and the STOMP broker from another origin with tokens from the configured identity provider.
Things that did not work on the cognito and keycloak profiles:
- no CORS
- preflights fell into the login redirect
- unauthenticated XHR got a 302
- a bearer token yielded a different identity than the cookie session of the same person
- STOMP was same-origin and cookie-only

### What does this PR do?

This adds the following:
- CORS from `DIRIGIBLE_CORS_*` (unset keeps the basic/snowflake wildcard, now without credentials).
- A 401 with `WWW-Authenticate: Bearer` and a JSON body for programmatic requests on the OAuth2 chains. Navigations still redirect.
- **Kind-aware bearer validation**: ID tokens are verified like a login (issuer, audience, principal claim, verified e-mail, group roles); access tokens keep signature/expiry/scope roles plus the issuer check. `DIRIGIBLE_OAUTH2_JWT_*` overrides the defaults.
- **STOMP**: configured origins on the handshake, `Authorization: Bearer` on CONNECT, and frame authorization (CONNECT needs a principal, SUBSCRIBE only `/user/queue/**`, SEND only `/ws/**`).
- `POST /login/token`: exchanges a bearer ID token for the platform session, which ends with the token.
- **A Keycloak-profile token without `typ` is refused**: Keycloak types every token it issues, so an untyped one is another issuer's.
- Accepting access tokens without `DIRIGIBLE_OAUTH2_JWT_AUDIENCES` logs a WARN at start. Every access token of the issuer is a credential here.
- **Tenant selection for bearer requests under `TOKEN_GROUPS`** (#7446): a bearer request names its tenant in the `X-Tenant-Id` header, validated against the token's groups exactly as a session selection is (403 `NOT_A_MEMBER`, 409 `NOT_PROVISIONED_HERE` / `UNKNOWN_HERE`), and is served in that tenant with the roles its groups grant there. Omitting the header results in a default tenant, global roles only, an ID token with no global role is refused (403) - access tokens untouched. `X-Tenant-Id` is a default CORS allowed header.
- The platform's own `websockets` API client is refused by a Dirigible `/stomp` unless it passes `Authorization: Bearer <token>` in the new `connectHeaders` argument.
- A `/login/token` session is not re-validated at the identity provider before the ID token's `exp`.
- The STOMP endpoint is left out of the platform CORS filter. `DIRIGIBLE_CORS_ALLOW_CREDENTIALS` does not apply to it.
- A wildcard or scheme-less origin pattern in `DIRIGIBLE_CORS_ALLOWED_ORIGINS` no longer reaches the STOMP handshake, since it carries the session cookie regardless of `DIRIGIBLE_CORS_ALLOW_CREDENTIALS`. Only patterns naming a host are applied there, with a boot warning for any dropped.
- The bearer decoder now accepts every asymmetric JWS algorithm (RS/PS/ES) instead of RS256 only, and fetches the JWKS through Nimbus with caching, rate limiting, and bounded timeouts instead of Spring's uncapped default.

All of the above are profile-agnostic.

#### Release Notes

Behavior changes without configuration:
- An ID token used as a bearer now acts as the user named by the principal claim with the roles of its groups (before: `sub`, no roles) and is refused for a foreign `aud`, a missing principal claim, an unverified `email` name, an unexpected Keycloak `typ`, or a mismatched `iss`. Access tokens now require the profile's issuer.
- basic/snowflake CORS no longer sends credentials cross-site.
- STOMP requires an authenticated CONNECT. Аnonymous clients and `/topic` subscriptions stop.
- OAuth2 chains no longer create a session per anonymous or bearer request.
- In single-pool/realm multi-tenant mode an ID token must carry `custom:tenant`. Аccess tokens without it pass as before.
- The "no global role" refusal of an ID token under TOKEN_GROUPS moves from 401 (token refused) to 403 (request refused), and `DIRIGIBLE_CORS_ALLOWED_HEADERS` gains `X-Tenant-Id` by default.

#### What is verified
- unit suites of the changed modules
- `formatter:validate`
- the release javadoc profile
- `ExternalFrontendIT` (30 tests, keycloak profile against an in-process OIDC provider: CORS, 401 shapes, ID/access tokens, hosted login, native login, STOMP, `/login/token`) and `EndpointAuthorizationIT` (5 tests), all green locally on H2
- Added `ExternalFrontendTenantSelectionIT` (8 cases, `TOKEN_GROUPS`, keycloak profile against the in-process OIDC provider), the two new `ExternalFrontendIT` cases (same-origin write with `Origin`, proxy failure mode), `TenantSelectionIT` and `TokenGroupsTenantResolutionIT`.
- `StompWildcardOriginIT` (3 cases) and a new `ExternalFrontendIT` case for the cookie-authenticated CONNECT (30 cases total now)
- Unit suites of the six touched modules, formatter validate, release javadoc profile, docs site build

#### What was NOT verified
- End-to-end for PostgreSQL and the Cognito profile.

#### Documentation

dirigible-io/dirigible-io.github.io (new page `external-frontends.md`, env-var tables, see-also links), opened as a separate PR (https://github.com/dirigible-io/dirigible-io.github.io/pull/249).

Resolves #7415 and #7446


components/engine/engine-websockets/src/main/java/org/eclipse/dirigible/components/websockets/config/BearerTokenStompInterceptor.java
```diff
@@ -0,0 +1,132 @@
...
+        if (sessionAttributes != null && authenticated.expiresAt() != null) {
+            sessionAttributes.put(EXPIRES_AT_ATTRIBUTE, authenticated.expiresAt());
+        }
+        LOGGER.debug("Authenticated the STOMP session [{}] of user [{}] by a bearer token", accessor.getSessionId(),
+                authenticated.authentication()
+                             .getName());
+        accessor.setUser(authenticated.authentication());
+    }
+
+    private static void refuseWhenExpired(Map<String, Object> sessionAttributes) {
```
iliyan-velichkov, line 146: **M1 (medium, correctness / resource leak): the synthetic DISCONNECT of an expired bearer session is refused here, so the broker never unregisters the session.**

When the WebSocket of a bearer session closes after the token's `exp`, Spring's `StompSubProtocolHandler.afterSessionEnded` (spring-websocket 7.0.8, line 670) synthesizes a `DISCONNECT` carrying the session attributes and sends it through `clientInboundChannel` so that `SimpleBrokerMessageHandler` drops the session from its `DefaultSubscriptionRegistry`. `preSend` treats that DISCONNECT like any other non-CONNECT frame, `refuseWhenExpired` finds `EXPIRES_AT_ATTRIBUTE` in the attributes and throws. `afterSessionEnded` has no catch around `outputChannel.send` (only a `finally`), so the exception propagates to `ExceptionWebSocketHandlerDecorator.afterConnectionClosed`, which logs "Unhandled exception after connection closed" at WARN with the stack trace, and the DISCONNECT never reaches the broker.

Effect: the session's subscriptions stay in the broker registry for the life of the process, and any token holder can grow it (connect, subscribe, wait for `exp`, close), plus one stack trace per expired session.

Fix: exempt `StompCommand.DISCONNECT` from the expiry check (a client leaving must always be allowed to leave), and add a unit test with a DISCONNECT frame whose session attributes carry a past `EXPIRES_AT_ATTRIBUTE`. `framesOfASessionWhoseTokenExpiredAreRefused` covers SUBSCRIBE and SEND only.

components/engine/engine-websockets/src/main/java/org/eclipse/dirigible/components/websockets/config/BearerTokenStompInterceptor.java
```diff
@@ -0,0 +1,132 @@
...
+        }
+        StompCommand command = accessor.getCommand();
+        if (command == StompCommand.CONNECT || command == StompCommand.STOMP) {
+            String authorization = accessor.getFirstNativeHeader(HttpHeaders.AUTHORIZATION);
+            if (authorization != null) {
+                authenticate(accessor, authorization);
+            }
+            return message;
+        }
+        refuseWhenExpired(accessor.getSessionAttributes());
```
iliyan-velichkov, line 105: **M2 (medium): expiry is enforced on inbound frames only, so the "connection lives no longer than the token" claim (class Javadoc, and `external-frontends.md`: "A session opened with a token ends when the token expires") does not hold.**

Heartbeats do not traverse the inbound channel and outbound delivery is never checked. A client that subscribed to `/user/queue/**` and then sends nothing keeps receiving `convertAndSendToUser` messages indefinitely after `exp`, until it sends a frame or the socket dies.

Either schedule a close at `expiresAt` (record `sessionId -> expiresAt` on CONNECT and close the `WebSocketSession` with `CloseStatus.POLICY_VIOLATION` from a `TaskScheduler`, or check the attribute in an interceptor on `clientOutboundChannel` and close there), or re-word the Javadoc and the docs to "any frame the client sends after expiry is refused".

components/security/security-oauth2/src/main/java/org/eclipse/dirigible/components/security/oauth2/resourceserver/TokenKind.java
```diff
@@ -0,0 +1,119 @@
...
+        if ("access".equals(tokenUse)) {
+            return Optional.of(ACCESS);
+        }
+        return Optional.empty();
+    }
+
+    private static Optional<TokenKind> ofKeycloak(String type) {
+        // a Keycloak access token is typed Bearer; a token typed nothing at all is read as one, so a
+        // realm whose mappers drop the claim keeps the access-token rules that applied before
+        if (type == null || "Bearer".equals(type)) {
```
iliyan-velichkov: **M3 (medium, validation gap): a token without `typ` is classified as an access token.**

`DIRIGIBLE_OAUTH2_JWT_ISSUER_URI` / `DIRIGIBLE_OAUTH2_JWT_JWK_SET_URI` exist so the profile can be pointed at another issuer (the IT does exactly that with a mock). Most OIDC providers (Auth0, Okta, Entra ID, Spring Authorization Server) do not put `typ` in the payload. Their ID tokens therefore land on the ACCESS branch: no audience check unless `DIRIGIBLE_OAUTH2_JWT_AUDIENCES` is set, `sub` as the user name, roles from scopes only, and in `KeycloakTenantFilter` exempt from the `custom:tenant` membership check that ID tokens must pass (`idToken || tenantClaim != null`). Nothing is granted that an access token would not get, but the deployment silently runs the less-validated path while the docs promise ID tokens are "verified like a login".

Suggestion: treat an absent `typ` as unknown (refused) behind an explicit opt-in such as `DIRIGIBLE_OAUTH2_JWT_UNTYPED_TOKEN_KIND=access` for realms that really strip the claim; at minimum refuse to start when `ACCESS` is accepted, no audiences are configured and the issuer was overridden away from the profile's own.

components/security/security-oauth2/src/main/java/org/eclipse/dirigible/components/security/oauth2/resourceserver/ResourceServerJwtSettings.java
```diff
@@ -0,0 +1,199 @@
...
+                    DirigibleConfig.OAUTH2_JWT_AUDIENCES.getKey());
+        }
+        Set<String> idTokenAudiences = configuredAudiences.isEmpty() ? audienceOf(clientId) : Set.copyOf(configuredAudiences);
+        if (idTokenAudiences.isEmpty()) {
+            throw new InvalidConfigException(
+                    "The audience of bearer ID tokens is unknown: the login profile has no client id and ["
+                            + DirigibleConfig.OAUTH2_JWT_AUDIENCES.getKey() + "] names none",
+                    DirigibleConfig.OAUTH2_JWT_AUDIENCES.getKey());
+        }
+        Set<String> accessTokenAudiences = audiencesConfigured ? Set.copyOf(configuredAudiences) : Set.of();
```
iliyan-velichkov, line 137: **M4 (medium, pre-existing, now issuer-scoped): access tokens are not audience-checked unless configured.**

Any RS256 token of the configured issuer marked `typ=Bearer` (Keycloak) or `token_use=access` (Cognito) is accepted regardless of the client it was issued to, and its `scope` / `scp` values are mapped to roles through the `ScopeService` mappings. A token minted for an unrelated API in the same pool or realm is a valid credential here if its scopes match a mapping; in trial mode `ScopeRoleJwtAuthoritiesConverter` grants every role to any such token. This was the case before the PR too, and the PR adds the issuer check and documents the gap honestly.

A tighter default is not safe (Cognito machine clients use a different app client than the login client; Keycloak access tokens carry `aud=account` unless a mapper is added), so: a boot-time WARN when `ACCESS` is an accepted kind and `accessTokenAudiences` is empty, and a sentence in the docs' "Send the right token" section telling machine-to-machine deployments to set `DIRIGIBLE_OAUTH2_JWT_AUDIENCES`.

components/core/core-base/src/main/java/org/eclipse/dirigible/components/base/http/access/CorsConfigurationSourceProvider.java
```diff
@@ -9,30 +9,228 @@
...
+            LOGGER.warn("Trial mode grants every role to every user, and the origins {} may now call the platform cross-origin.", origins);
+        }
+    }
+
+    /**
+     * Whether an origin pattern matches every origin: the bare wildcard, or a pattern whose host is
+     * nothing but the wildcard ({@code https://*}, {@code *://*:8080}) - as good as the bare one for
+     * whoever wants to reach the platform from a page of their own.
+     */
+    static boolean matchesEveryOrigin(String origin) {
```
iliyan-velichkov, line 241: **M5 (medium, guard bypass through operator configuration): the "every origin with credentials" refusal misses Spring's port-list syntax.**

`allowedOriginPatterns` accept a port list (`https://*.example.com:[8080,8081]`, `https://*.example.com:[*]`). The pattern `https://*:[*]` (or `*://*:[*]`) matches every https origin on every port, but after the wildcard substitution it becomes `https://wildcard.invalid:[wildcard.invalid]`, which `new URI(...)` rejects; `parsePattern` returns `null`, this method returns `false`, and `refuseUnsafe` accepts credentials for it.

Fix: strip a trailing `:[...]` port list before parsing, and when credentials are enabled treat an unparseable pattern as unsafe (refuse) rather than as safe. Add `https://*:[*]` to `credentialsForEveryOriginAreRefused`.

components/core/core-tenants/src/main/java/org/eclipse/dirigible/components/tenants/tenant/TenantExtractor.java
```diff
@@ -95,10 +98,20 @@ public Optional<Tenant> determineTenant(HttpServletRequest request) {
...
+     * <p>
+     * A request authenticating with a bearer token lands in the default tenant as well, whatever
+     * session cookie rides along: the token proves an identity of its own, and running it in the tenant
+     * another identity selected would let a page holding both act across the two.
+     *
      * @param request the request
      * @return the selected tenant, or the default tenant
      */
     private Optional<Tenant> determineSelectedTenant(HttpServletRequest request) {
+        if (ProgrammaticRequestMatcher.isBearerAuthorization(request.getHeader(AUTHORIZATION_HEADER))) {
```
iliyan-velichkov, line 119: **M6 (medium, functional gap not stated in the PR description): under `TOKEN_GROUPS` a bearer request is always pinned to the default tenant.**

Together with `ResourceServerJwtSupport` refusing an ID token whose user has only tenant-scoped groups ("The token grants no role of this deployment"), an external SPA on a `TOKEN_GROUPS` deployment can act only for platform staff, which is the opposite of the multi-tenant frontend case #7415 describes. The docs PR says "Selecting a tenant on a bearer request is not supported yet"; the release notes here do not. Please state it, and track a follow-up (a header or claim driven selection validated against the token's groups, the way `TenantSelectionManager.selectTenant` validates a session selection).

Also worth a comment in the code: the tenant decision keys on the mere presence of an `Authorization: Bearer` header, before any authentication. It is safe on the cognito, keycloak and github chains because the resource-server filter answers 401 for an invalid token before anything else runs, but that safety depends on every chain that can run `TOKEN_GROUPS` having a resource server. Resolving the tenant from the authenticated principal type would remove the implicit coupling.

components/security/security-oauth2/src/main/java/org/eclipse/dirigible/components/security/oauth2/resourceserver/ResourceServerJwtSupport.java
```diff
@@ -0,0 +1,249 @@
...
+            TokenKind kind = settings.kindOf(jwt)
+                                     .orElseThrow(() -> new InvalidBearerTokenException(KIND_NOT_ACCEPTED.getDescription()));
+            return kind == TokenKind.ID ? idTokenConverter.convert(jwt) : accessTokenConverter.convert(jwt);
+        }
+
+        private Collection<GrantedAuthority> idTokenAuthorities(Jwt jwt,
+                Converter<Jwt, Collection<GrantedAuthority>> scopeAuthoritiesConverter,
+                TenantAwareAuthoritiesMapper groupAuthoritiesMapper) {
+            Set<GrantedAuthority> groupAuthorities = groupAuthoritiesMapper.authoritiesOf(jwt);
+            if (groupAuthorities.isEmpty() && TenantResolutionStrategy.fromConfiguration() == TenantResolutionStrategy.TOKEN_GROUPS) {
```
iliyan-velichkov, line 292: Part of M6: under `TOKEN_GROUPS` this refuses every user who has only tenant-scoped groups, so the bearer path serves platform staff only. Fine as an interim rule, but it needs to be in the PR description's behaviour changes, not only in the docs PR.

components/engine/engine-websockets/src/main/java/org/eclipse/dirigible/components/websockets/config/WebsocketConfig.java
```diff
@@ -64,9 +99,48 @@ public void configureMessageBroker(final MessageBrokerRegistry config) {
...
+    }
+
+    /**
+     * The rules every client frame is held to.
+     *
+     * @return the authorization manager of the inbound channel
+     */
+    static AuthorizationManager<Message<?>> inboundAuthorization() {
+        return MessageMatcherDelegatingAuthorizationManager.builder()
+                                                           .nullDestMatcher()
```
iliyan-velichkov, line 152: **L1 (low): every refused anonymous CONNECT produces a WARN stack trace, from an unauthenticated public endpoint.**

`/stomp/**` is `permitAll`. For an anonymous handshake whose CONNECT is denied, Spring synthesizes a DISCONNECT without a user; `nullDestMatcher().authenticated()` denies that too, and the same "Unhandled exception after connection closed" WARN with stack trace is logged (same mechanism as M1), once per attempt.

Consider `.simpTypeMatchers(SimpMessageType.DISCONNECT).permitAll()` ahead of `nullDestMatcher()`, the pattern in Spring Security's own messaging docs (the method exists on the 7.1.0 builder). A DISCONNECT needs no guard.

components/engine/engine-websockets/src/main/java/org/eclipse/dirigible/components/websockets/config/WebsocketConfig.java
```diff
@@ -64,9 +99,48 @@ public void configureMessageBroker(final MessageBrokerRegistry config) {
...
+        registry.setErrorHandler(new BearerTokenStompErrorHandler());
+        StompWebSocketEndpointRegistration endpoint = registry.addEndpoint("/stomp");
+        StompWebSocketEndpointRegistration sockJsEndpoint = registry.addEndpoint("/stomp");
+        if (CorsConfigurationSourceProvider.isConfigured()) {
+            List<String> origins = CorsConfigurationSourceProvider.allowedOriginPatterns();
+            LOGGER.info("The STOMP handshake accepts the origins {}.", origins);
+            String[] originPatterns = origins.toArray(String[]::new);
+            endpoint.setAllowedOriginPatterns(originPatterns);
+            // the SockJS registration copies the patterns when it is created, so they go on first
+            sockJsEndpoint.setAllowedOriginPatterns(originPatterns);
```
iliyan-velichkov, line 124: **L2 (low): SockJS fallback transports cannot work cross-origin with the default `DIRIGIBLE_CORS_ALLOW_CREDENTIALS=false`.**

SockJS XHR transports send `withCredentials` and require `Access-Control-Allow-Credentials: true`. The chain-level `CorsFilter` (registered for `/**`) answers `/stomp/**` requests first without that header, and Spring's `DefaultCorsProcessor` skips the SockJS service's own CORS configuration when `Access-Control-Allow-Origin` is already present ("Skip: response already contains Access-Control-Allow-Origin"). Only the raw WebSocket transport, which the docs example uses, works from another origin.

Document it next to the STOMP example, or exclude `/stomp/**` from the chain CORS source and let the SockJS service handle its own origins.

components/security/security-oauth2/src/main/java/org/eclipse/dirigible/components/security/oauth2/login/NativeLoginSessionInitializer.java
```diff
@@ -106,16 +118,73 @@ void establishSession(ClientRegistration registration, NativeLoginTokens tokens,
...
+        OAuth2AuthenticationToken authentication = new OAuth2AuthenticationToken(oidcUser, authorities, registrationId);
+        authentication.setDetails(new WebAuthenticationDetailsSource().buildDetails(request));
+
+        HttpSession carriedSession = request.getSession(false);
+        if (carriedSession != null) {
+            carriedSession.invalidate();
+        }
+        HttpSession session = request.getSession(true);
+        saveContext(authentication, request, response);
+        session.setAttribute(OAuth2SessionRevalidationFilter.SESSION_EXPIRES_AT_ATTRIBUTE, jwt.getExpiresAt());
```
iliyan-velichkov, line 163: **L3 (low): a `/login/token` session is never re-validated against the identity provider.**

A hosted-login session is re-validated through the refresh token whenever the access token expires, so revocation at the IdP ends it within the access-token lifetime. A token-minted session has no refresh token and lives until the ID token's `exp` (Cognito ID-token validity is configurable up to 24 hours). A user disabled at the IdP keeps such a session until then. Acceptable for bearer semantics, but say so in the docs' "Exchange a token for a session" section, and consider a platform maximum.

tests/tests-integrations/src/main/java/org/eclipse/dirigible/integration/tests/api/ExternalFrontendIT.java
```diff
@@ -0,0 +1,863 @@
...
+    @Autowired
+    private SimpMessagingTemplate messagingTemplate;
+
+    @Autowired
+    private SimpUserRegistry simpUserRegistry;
+
+    @BeforeAll
+    static void startIdentityProviderAndAllowTheOrigin() throws Exception {
+        identityProvider = MockIdentityProvider.start();
+        Configuration.set(DirigibleConfig.CORS_ALLOWED_ORIGINS.getKey(), ORIGIN);
```
iliyan-velichkov, line 167: **L4 (low, test hygiene): this static configuration is never removed.**

`Configuration` is JVM-static and the `api` shard runs many IT classes per fork, so every later IT in the fork boots with CORS configured for `https://app.example.com` (the basic chain then answers that origin instead of `*`, `CorsSecurityConfigurator` logs). Nothing asserts on it today, so no visible failure, only a hidden coupling. Add `Configuration.remove(DirigibleConfig.CORS_ALLOWED_ORIGINS.getKey())` in `@AfterAll`, as the unit tests do in `clearConfiguration`.

components/security/security-cognito/src/main/java/org/eclipse/dirigible/components/security/cognito/CognitoTenantFilter.java
```diff
@@ -77,30 +84,45 @@ protected void doFilterInternal(HttpServletRequest request, HttpServletResponse
...
+                if ((idToken || tenantClaim != null) && !isMember(tenantClaim, currentTenant.get(), response)) {
                     return;
                 }
             }
         }
         chain.doFilter(request, response);
     }
 
+    private boolean isMember(String tenantAttribute, Tenant currentTenant, HttpServletResponse response) throws IOException {
+        if (tenantAttribute == null || tenantAttribute.equals("")) {
```
iliyan-velichkov: Nit (CodeQL): `tenantAttribute.isEmpty()`.

components/security/security-keycloak/src/main/java/org/eclipse/dirigible/components/security/keycloak/KeycloakTenantFilter.java
```diff
@@ -75,30 +82,45 @@ protected void doFilterInternal(HttpServletRequest request, HttpServletResponse
...
+                if ((idToken || tenantClaim != null) && !isMember(tenantClaim, currentTenant.get(), response)) {
                     return;
                 }
             }
         }
         chain.doFilter(request, response);
     }
 
+    private boolean isMember(String tenantAttribute, Tenant currentTenant, HttpServletResponse response) throws IOException {
+        if (tenantAttribute == null || tenantAttribute.equals("")) {
```
iliyan-velichkov: Nit (CodeQL): `tenantAttribute.isEmpty()`.

iliyan-velichkov (changes requested):
## Security review

Reviewed at `725156c55b` by reading the complete production diff, both IT classes and the tests, plus the unchanged code the change depends on (`HttpSecurityURIConfigurator`, `SecurityFilter`, the tenant selection classes, `UserFacade`, `ScopeRoleJwtAuthoritiesConverter`, `WebsocketController`, `WebsocketClient`, the profile properties) and Spring's `StompSubProtocolHandler` / `ExceptionWebSocketHandlerDecorator` / `DefaultCorsProcessor` sources where a finding depends on framework behaviour. The unit suites of the six changed modules were run locally (317 tests, all green).

**Verdict.** Net security improvement with correct core token handling. It closes two real pre-existing holes: the STOMP broker had no authorization at all on a public endpoint (anonymous subscribe to any `/topic` or `/queue`, including other sessions' `-user<sid>` queues, and send to `/ws/**`), and the basic and snowflake chains served wildcard CORS with credentials while CSRF tokens are disabled on every chain. The new bearer path validates issuer, expiry, audience, principal claim and e-mail verification consistently between the HTTP chain and the STOMP CONNECT, bearer requests create no session, and `POST /login/token` mints only from a validated ID token into a fresh session. I found no introduced vulnerability that gives a caller access it did not have before.

| # | Severity | Where | One line |
|---|----------|-------|----------|
| M1 | Medium, correctness | `BearerTokenStompInterceptor` | Synthetic DISCONNECT of an expired bearer session is refused: broker subscriptions leak, WARN stack trace per session |
| M2 | Medium, claim not implemented | `BearerTokenStompInterceptor` | Expiry enforced on inbound frames only; a silent subscriber keeps receiving after `exp` |
| M3 | Medium, validation gap | `TokenKind` | Keycloak profile treats a token without `typ` as an access token |
| M4 | Medium, pre-existing | `ResourceServerJwtSettings` | Access tokens not audience-checked unless configured |
| M5 | Medium, guard bypass | `CorsConfigurationSourceProvider` | "Every origin + credentials" refusal misses `https://*:[*]` |
| M6 | Medium, undocumented gap | `TenantExtractor`, `ResourceServerJwtSupport` | `TOKEN_GROUPS`: bearer requests pinned to the default tenant, tenant-only users refused |
| L1 | Low | `WebsocketConfig` | Refused anonymous CONNECTs log a WARN stack trace each |
| L2 | Low | `WebsocketConfig` | SockJS fallbacks cannot work cross-origin with credentials off |
| L3 | Low | `NativeLoginSessionInitializer` | `/login/token` sessions never re-validated at the IdP until `exp` |
| L4 | Low | `ExternalFrontendIT` | Static CORS config leaks into later IT classes |
| L5 | Low | release notes | The platform's own `websockets` API client is now refused at CONNECT |
| L6 | Info | CodeQL | 13 alerts, none real; two cosmetic fixes |

M1 to L4 are inline. Requesting changes for M1 (one-line fix plus a test) and M2 (implement or re-word); M3 to M6 in this PR or as tracked follow-ups.

### L5. Behaviour change missing from the release notes
`components/api/api-net/.../WebsocketClient.java` (the server-side `websockets` JS API) connects with `WebSocketStompClient.connectAsync(uri, handler)` and no credentials. User code that uses it against a Dirigible `/stomp` (its own instance or another one) is now refused at CONNECT with `Unauthorized`. The notes say "anonymous clients stop" but do not name the platform's own client; please add it here and to the docs' "What changes for existing deployments" list.

### L6. CodeQL alerts (13)
- Log injection (9): every flagged value is operator configuration read through `DirigibleConfig` (origins, issuer, audiences), not request input. Dismiss with that justification; stripping CR/LF before logging is a cheap defensive option.
- Sensitive information into log files (`ResourceServerJwtSettings` 149): the INFO line logs issuer, audiences, kinds and the principal claim name. None is a secret. Dismiss.
- Inefficient empty string test (both tenant filters): inline nits.
- Useless parameter `reason` (`ExternalFrontendIT`): it is the `@ParameterizedTest(name = "{0}")` label. Dismiss.

### Reading notes, no action required
- `ProgrammaticRequestMatcher.matches`: `Sec-Fetch-Mode: navigate` short-circuits before the bearer check. Harmless, because `OAuth2ResourceServerConfigurer` registers the same entry point for its own `BearerTokenRequestMatcher`; a one-line comment would save the next reader the analysis.
- There are now three definitions of "programmatic request": `ProgrammaticRequestMatcher`, `TenantSelectionFilter.prefersHtml`, `TenantContextInitFilter.prefersHtml`.
- `BearerUnauthorizedEntryPoint.challengeOf` sanitizes `error_description` for the header by replacing `"` only; the descriptions are Spring and Nimbus messages without CR/LF, same posture as Spring's own entry point.

### Verified as correct (so nobody redoes it)
- Entry point ordering on the OAuth2 chains: the explicit `defaultAuthenticationEntryPointFor` registration precedes the configurers' `init`, so it is the first matcher tried and the `DelegatingAuthenticationEntryPoint` default. Navigations redirect, scripts and header-less clients get the 401 JSON.
- ID-token validation: `createDefaultWithIssuer` (timestamps, `nbf`, issuer) on both kind branches, plus `exp`, `aud`, principal claim and `email_verified` for ID tokens; kind read identically in validator and converter; RS256-only decoder; the STOMP `authenticate()` runs a `JwtAuthenticationProvider` over the same decoder and converter.
- Roles: `authoritiesOf(Jwt)` shares `mapGroups` with the login mapper; `ROLE_` prefix via `AuthoritiesUtil`, so `request.isUserInRole`, `UserFacade`, the `.access` `SecurityFilter` and the URL role gates all see them; `UserFacade.getName()` resolves to the principal claim.
- No session for bearer requests (`RequestAttributeSecurityContextRepository` + `IF_REQUIRED`); hosted and native login still create theirs.
- `/login/token`: only a validated ID-kind `JwtAuthenticationToken` mints; cookie-only 401, access token 403, no support 404; carried session invalidated, fresh session, `FACTOR_*` stripped, `exp` enforced by the revalidation filter, authorized-client store untouched on purpose. A browser cannot attach `Authorization` cross-origin without a passing preflight.
- CORS: `null` origin refused, wildcard with credentials refused (modulo M5), `*` headers with credentials refused, unsupported methods refused; `CorsFilter` precedes authentication, `AuthorizationFilter` and the servlet-level `.access`, tenant and tenant-selection filters, so preflights never reach them; `CorsSecurityConfigurator` correctly skips chains that already have a `CorsConfigurer`.
- STOMP: CONNECT needs a principal, SUBSCRIBE only `/user/queue/**` (rewritten to the caller's own session), SEND only `/ws/**`, everything else denied; the bearer user is set while the CONNECT accessor is still mutable so Spring's user callback records it for the session; ERROR frames and `WebsocketController` no longer echo exception text; the only platform destination pair is inside the allow-list and no shipped JS/TS client subscribes to `/topic`.

Not run here: `ExternalFrontendIT`, `formatter:validate`, the release javadoc profile, the Cognito profile end to end, the PostgreSQL leg.

Local build note: `spring-security-messaging:7.1.0` is new, so an offline (`-o`) build fails at `engine-websockets` until it has been downloaded once.

components/engine/engine-websockets/src/main/java/org/eclipse/dirigible/components/websockets/config/BearerTokenStompInterceptor.java
iliyan-velichkov: Fixed in edae828def: a `DISCONNECT` is never refused any more, so the synthetic one Spring sends on close reaches the broker and the session's subscriptions are dropped. `aDisconnectIsNeverRefusedAndEndsTheDeadline` covers the frame; in `ExternalFrontendIT` the expired session that the terminator closes no longer produces the "Unhandled exception after connection closed" warning (the two that remain in that run both follow a refused CONNECT, i.e. L1).

components/engine/engine-websockets/src/main/java/org/eclipse/dirigible/components/websockets/config/BearerTokenStompInterceptor.java
iliyan-velichkov: Fixed in edae828def: new `BearerTokenStompSessionTerminator` (a `@Component`, resolved lazily from the configurer to stay clear of a cycle with the broker beans). On an accepted bearer CONNECT it schedules an ERROR frame reading `Unauthorized` on `messageBrokerTaskScheduler` for the token's `exp`; Spring closes the WebSocket after sending an ERROR frame, and a DISCONNECT before the deadline cancels it. The inbound expiry check stays as the second line. `BearerTokenStompSessionTerminatorTest` (5 cases) and the new IT case `aBearerStompSessionEndsWhenItsTokenDoes` (a subscribed client that sends nothing is closed at `exp`, 25/25 green locally) cover it. The Javadoc and the docs claim now hold as written.

components/security/security-oauth2/src/main/java/org/eclipse/dirigible/components/security/oauth2/resourceserver/TokenKind.java
iliyan-velichkov: Fixed in a341903e37: `TokenKind.ofKeycloak` no longer reads a missing `typ` as an access token. Such a token has no kind and is refused, like a `Refresh` one. Keycloak types every token it issues, so nothing Keycloak-issued changes; the ID token of another issuer the profile is pointed at is refused instead of landing on the weaker rules (and can no longer slip past the tenant filter's ID-token membership check). Covered by `TokenKindTest` and a new Keycloak untyped case in `ResourceServerJwtSupportTest.aRefreshTokenAndATokenOfUnknownUseAreRefused`; `ExternalFrontendIT` re-run 25/25. The docs PR's "What changes" bullet says so now (dirigible-io/dirigible-io.github.io@316a687f2). I did not add the opt-in for realms that strip `typ`: no supported provider needs it, and it can come with a real case.

components/security/security-oauth2/src/main/java/org/eclipse/dirigible/components/security/oauth2/resourceserver/ResourceServerJwtSettings.java
iliyan-velichkov: Fixed in a341903e37 as proposed: `ResourceServerJwtSettings.of` logs a WARN at start when `access` is an accepted kind and `DIRIGIBLE_OAUTH2_JWT_AUDIENCES` is unset, naming the key and what it means. The default itself stays (Cognito machine clients use a separate app client, Keycloak access tokens carry `aud=account`), and the docs' "Send the right token" bullet now tells a deployment that accepts access tokens to set it (docs PR branch, 316a687f2).

components/core/core-base/src/main/java/org/eclipse/dirigible/components/base/http/access/CorsConfigurationSourceProvider.java
iliyan-velichkov: Fixed in a341903e37: `parsePattern` drops a trailing port pattern (`:[8080]`, `:[*]`, `:*`) before parsing, so `https://*:[*]`, `*://*:[*]` and `https://*:*` are recognised as every-origin and refused with credentials; and with credentials enabled a pattern that still cannot be parsed is refused as uncheckable (under `DIRIGIBLE_CORS_ALLOWED_ORIGINS`) instead of being trusted. `CorsConfigurationSourceProviderTest` covers the every-origin variants, a narrow `https://*.example.com:[8080]` with credentials (still allowed), an unparseable pattern with credentials (refused) and without (allowed).

One thing the test surfaced: the value is split on commas by `getListValue()`, so Spring's multi-port list `:[8080,8081]` cannot be configured through this key at all. It arrives as `https://*.example.com:[8080` and `8081]`, two patterns that match nothing, and with credentials on they are now refused as uncheckable. Single-port lists and `:[*]` work. Worth a sentence on the env-var page, or a bracket-aware split if multi-port lists are wanted.

StanZGenchev:
* L1
Should be fixed. DISCONNECT is "permitAll" ahead of `nullDestMatcher();`. `WebsocketChannelSecurityTest.aDisconnectIsNeverRefused` pins it, and the "Unhandled exception after connection closed" lines are gone from the ExternalFrontendIT run.

* L2
Fixed it by following your proposal. `/stomp/**` is left out of the chain CORS source. The endpoint checks origins against the same list and SockJS answers its transports with credentials. This is not a widening. The raw WebSocket handshake already carried cookies for those origins, and before the PR, the endpoint was same-origin only. For the integration tests, `/stomp/info` from the listed origin carries `Access-Control-Allow-Credentials: true`, from an unlisted one 403.

* L3
Documented in "Exchange a token for a session". No platform maximum. The same token is accepted as a bearer for the same span, so a cap on the session alone would protect nothing.

* L4
This was not true. `IntegrationTest.reloadConfigurations() (tests-framework, @AfterAll)` calls `Configuration.reloadConfigurations()`, whose first line is:
```js
RUNTIME_VARIABLES.clear();
```
Superclass `@AfterAll` runs after the subclass's, so the value is gone before the next IT class boots.

* L5
Fixed. `createWebsocket(uri, handler, connectHeaders)` on the facade, the JS module and the Java SDK. WebsocketClient sends them on the CONNECT frame. For te integration tests, the platform's client connects to its own `/stomp` with a bearer token and is refused without one. Release notes and docs name it.

* L6
Four alerts fixed in code (the port-pattern regex, both `isEmpty()`, Named replaces the reason parameter). The nine log-injection alerts and the sensitive-log one are dismissed as false positives.

* M6
Release notes state it, `TenantExtractor.determineSelectedTenant` explains why the header decides (the filter precedes authentication on the chains that add it, so no principal exists there).

components/core/core-tenants/src/main/java/org/eclipse/dirigible/components/tenants/tenant/TenantExtractor.java
StanZGenchev: Opened issue #7446 for this as a follow-up. This PR is already too big.

Issue #7289, opened 2026-09-10
intent: notifications/integrations/outbound event.when never resolves a seeded status name and degrades an unparsable guard to true


delchev:
## Describe the bug

Surfaced while checking #7251 / PR #7269 for other `where`-shaped sites. The `event.when` guard of `notifications[]`, `integrations[]` and `outbound[]` is a status-guard site that neither resolves a seeded status name nor checks the guard's shape:

- `StatusSymbolResolver.resolve` (`parser/StatusSymbolResolver.java:82-89`) rewrites entities, transitions, processes, postings, generates, schedules, resolves and reports - nothing for the three glue lists.
- `IntentParser.validateEventBinding` (`:2666-2722`) / `validateNotifications` (`:2632-2665`) contain no `when` handling at all.
- The renderer, `NotificationSupport.guard` + `literalToJava` (`generator/NotificationSupport.java:296-342`): an unquoted word that is not a number/boolean falls through to `return quote(rhs);`.

So the most natural authoring of the construct

```yaml
notifications:
  - name: issued-mail
    event: { onUpdate: SalesInvoice, when: "Status == ISSUED" }
```

generates `if (!(java.util.Objects.equals(entity.Status, "ISSUED"))) return;` - a `Long` compared with a `String` - and the mail never goes out, with parse, generation, compile and publish all green. Every other guard in the DSL (`transitions`, `abortOn.status`, `wait when`, `resolves event.when`, posting `event.when`, `requiredWhen`, `generates event.when`, reports `filter`, and since #7269 `schedules[].where`) accepts the seeded name; #6645's summary in `intent-layer.md` ("every site that names a status additionally accepts the seeded name") does not hold for this axis.

Second half of the same gap: `guard()` returns `"true"` for anything that does not match its one regex, so a guard with a typo (`Status = ISSUED`, `status == 'ISSUED' and ...`) silently disables itself and the notification fires on every update - the degradation #7094 (`requiredWhen`) explicitly refused to accept for its own `when`.

## Expected

- `event.when` on `notifications`/`integrations`/`outbound` is rewritten by `StatusSymbolResolver` on the event entity's nomenclature, as `resolves.event.when` and posting `event.when` already are.
- A `when` that does not parse is a parse-time refusal, not `true` (the `requiredWhen` grammar and its typed-literal check are the model).

PR #7447, opened 2026-09-21
docs(intent): make the lookup-outcome examples show the shape the prose asks for

TIVMOF:
## The defect

#7414 (mine) corrected the prose about telling a lookup's two failure outcomes apart, and **left both worked examples showing the shape it had just stopped recommending.**

Prose, line 3937:

> Prefer that when the two failures are handled the SAME way and differ only in what the trail records: **one status, one fallback process, one worklist.**

Example, fifteen lines below:

```yaml
    event: { onTransition: Fine, when: "Status == NO_MATCH", mode: append }
processes:
  - name: ResolveNoMatch          # a fallback process per failure kind:
```

## Why it mattered

An intent generated from a full process description afterwards contained a process named **`ResolveNoMatch`** — a string that occurs exactly once in the 4195-line guide, in that example — alongside the two-status split and the per-failure-kind processes. In the same file it used the outcome guard the corrected prose describes, for its success row.

The example was followed; the prose was not. That is the expected weighting: a worked example is a demonstration, prose beside it is a caveat.

Both generated processes were identical but for their trigger status and step name — exactly the case the prose says should be one status and one worklist.

## What changed

- **The canonical `resolves:` block** routes both failures to one `UNRESOLVED`, with an inline comment saying `outcome:` is what separates them.
- **The worked example** mints one audit row per failure kind from the shared status plus an outcome term, and **one** fallback process for both. The `# a fallback process per failure kind` comment — the most instruction-like line in the passage — is gone.
- The two-status split keeps its place as the **different-flows** case, with *"two processes identical but for their trigger"* named as the sign the wrong branch was taken.

Three things in the surrounding prose were tightened against the parser while rewriting, rather than restated:

- a `generates` guard takes **one** status term (a second numeric comparison is refused by name);
- its other terms are over **string/text** fields, not just string;
- the **`postings:` exception is promoted out of a parenthesis into its own paragraph** — a postings guard cannot read the trace at all, so a posting per outcome still needs a status per outcome whatever the flow. That is the one consumer the recommendation does not reach, and burying it in brackets was how the previous round went wrong.

## Verification

Against a running instance, not read off the source:

| check | result |
|---|---|
| the rewritten worked example, as a complete intent | `issues: []` |
| a `notifications` guard on `["Status == UNRESOLVED", "resolution == ambiguous"]` | `issues: []` — licenses naming notifications/integrations/outbound beside generates |
| two status terms in one `generates` guard | refused: *"event when declares more than one numeric comparison"* — licenses "one status term" |
| `warnIfGuardFieldIsEditable` exists and fires on a non-`readOnly` guard field | confirmed in `GlueIntentGenerator` |

And, because the previous two rounds were reasoned rather than observed — **the actual end-to-end test**: the same process description fed to a patched instance now produces

```yaml
processes:
  - name: ManualIdentification
    trigger: { onTransition: Fine, when: "status == UNRESOLVED" }
```

one process, `notFound`/`ambiguous` both routed to `UNRESOLVED`, and the whole intent validates with no issues.

Documentation only; no code changes.


TIVMOF:
Three end-to-end generation runs have now finished against a patched instance, same process description each time:

| run | processes | validation | `requiredWhen` gate |
|---|---|---|---|
| 1 | 1 — `ManualIdentification` | `issues: []` | present |
| 2 | 1 — `ManualIdentification` | `issues: []` | present |
| 3 | 1 — `ManualIdentification` | `issues: []` | present |

All three route `notFound` and `ambiguous` to a single `UNRESOLVED` and separate the audit rows on the outcome term. Before this change the same description produced two near-identical fallback processes, one named `ResolveNoMatch` after the example.

Unplanned side effect worth noting: all three also emitted the conditional check documented in #7414, which no earlier run had produced —

```yaml
- { kind: requiredWhen, field: driver, when: "status == DRIVER_IDENTIFIED",
    status: DRIVER_IDENTIFIED, message: "A fine cannot be marked identified without a driver." }
```

That is the rule stopping a task completion marking a fine identified with the driver still empty. I would not claim this change caused it — three runs is a small sample and the kind was already documented — but it did not appear in any run before the examples were aligned.

components/engine/engine-intent/src/main/resources/intent-assistant-guide.md
```diff
@@ -3920,42 +3920,58 @@ candidates, because a silently-wrong driver (or price, or approver) is worse tha
...
-```
+**Mint a status per outcome when the outcomes are handled DIFFERENTLY** - different people, a different
+form, a different flow. A process binds at most one trigger and a guard list is ANDed, never ORed, so
+two failure kinds that genuinely need two flows do need two statuses to start them. Two processes
+identical but for their trigger are the sign this was the wrong branch.
+
+**And mint one per outcome for a `postings:` reaction whatever the flow** - a posting's guard is the
+status term alone, no list and no string term, so it cannot read the trace. This is the one consumer
+the paragraph above does not apply to; every other reaction to a lookup's outcomes (`generates:`,
+`notifications:`, `integrations:`, `outbound:`) can read it.
```
delchev, line 3938: **"every other reaction ... can read it" is only true for `generates:`.**

Telling the two failures apart while keeping one status requires the ANDed list `when` (status term + trace term), and the list form is implemented only for `generates` events and process triggers:

- `notifications:` — `NotificationSupport.plan` calls `when.toString()` on the raw object, so a list arrives as `"[Status == 7, resolution == notFound]"`, fails `SIMPLE_COMPARISON`, and `guard()` returns `"true"`.
- `integrations:` / `outbound:` — `GlueIntentGenerator.guardFields` (line 2300, used at 3956 and 4018) calls `stringArg(event, "when")` with the same effect, and additionally sets `hasGuard = false`.

Neither the parser nor the generator reports anything — there is no shape validation of a notification/integration/outbound `when` at all, which is why this PR's `issues: []` check passes. Concrete failure: an author follows this paragraph and writes

```yaml
notifications:
  - event: { onTransition: Fine, when: ["Status == UNRESOLVED", "resolution == ambiguous"] }
    to: ...
```

and gets an **unguarded** listener that mails on every `Fine` transition, including `IDENTIFIED`.

Secondary point: these three consumers are also absent from `StatusSymbolResolver`'s rewrite sites, so even a scalar `when: "Status == UNRESOLVED"` there stays symbolic and compiles to `Objects.equals(entity.Status, "UNRESOLVED")` — always false.

Suggest scoping the sentence to `generates:` (and process triggers), or saying explicitly that on the other three the trace term must stand alone.

components/engine/engine-intent/src/main/resources/intent-assistant-guide.md
```diff
@@ -3920,42 +3920,58 @@ candidates, because a silently-wrong driver (or price, or approver) is worse tha
...
-```yaml
-    notFound:  { setStatus: UNRESOLVED }
-    ambiguous: { setStatus: UNRESOLVED }
-# ...
-    event: { onTransition: Fine, mode: append, when: ["Status == UNRESOLVED", "resolution == notFound"] }
-```
+**Mint a status per outcome when the outcomes are handled DIFFERENTLY** - different people, a different
+form, a different flow. A process binds at most one trigger and a guard list is ANDed, never ORed, so
+two failure kinds that genuinely need two flows do need two statuses to start them. Two processes
+identical but for their trigger are the sign this was the wrong branch.
```
delchev, line 3933: **"two failure kinds that genuinely need two flows do need two statuses to start them" is not true of the platform as built.**

A process `trigger:` takes the ANDed list form — `TriggerSupport.triggerWhen` returns the raw object, `NotificationSupport.guard(Object)` renders a list, and `StatusSymbolResolver.rewriteProcesses` resolves the status symbol inside a list. So two *different* processes can both trigger on `Status == UNRESOLVED` and be separated by `"resolution == notFound"` vs `"resolution == ambiguous"`.

The stated reason ("a process binds at most one trigger and a guard list is ANDed, never ORed") correctly explains why *one* process cannot cover two statuses, but it does not support the conclusion drawn from it.

As written this is the one paragraph that still instructs a generator to mint a status per outcome, and it will do so for the common "two flows" reading — reintroducing exactly the two-status split the rest of the PR removes.

Suggested rewording: *different flows need two triggers, which can be two processes on the same status separated by the outcome term; mint a second status only when the flows must also be distinguishable to a `postings:` reaction or to the status-only consumers.*

delchev (commented):
Reviewed the rewrite hunk-by-hunk against `IntentParser`, `StatusSymbolResolver`, `GlueIntentGenerator`, `NotificationSupport`, `TriggerSupport` and `Resolve.java.template`. Most of it verifies:

- `StatusSymbolResolver.rewriteGenerates`/`rewriteProcesses` do rewrite a LIST `when` and leave a non-status term alone, so `["Status == UNRESOLVED", "resolution == notFound"]` resolves in both the `generates` event and the process trigger of the new example.
- `validateGeneratesWhen` does refuse a second numeric comparison and does accept `string`/`text` fields only — both tightened prose claims are accurate.
- The `postings:` claim is accurate (`IntentParser` refuses a list and requires `\w+ == \d+`).
- `warnIfGuardFieldIsEditable` exists and fires as described (`GlueIntentGenerator:1288`), so the `readOnly: true` advice is backed by a real warning.
- Ordering is safe: `Resolve.java.template` stamps `outcome:` in the first targeted write and sets the status second, so a `-transitioned` guard on the trace field can never read a stale outcome.

Two findings inline.

components/engine/engine-intent/src/main/resources/intent-assistant-guide.md
delchev: Fixed in 3af114e — the paragraph now says two flows need two **triggers**, with the two-processes-on-one-status list form spelled out, and the `postings:` paragraph states the positive rule (a posting that must tell the failures apart is the one reason to mint a status per outcome).

components/engine/engine-intent/src/main/resources/intent-assistant-guide.md
delchev: Retracting this one — I checked it against a stale checkout. On current `master` the list form reaches all three of these consumers, so the sentence as written is correct:

- `StatusSymbolResolver.rewriteGlue` (GLUE_LISTS = `notifications` / `integrations` / `outbound`, since #7289/#7317) rewrites their `event.when` through the list-aware `rewriteWhen`, so a status symbol there does resolve.
- `NotificationSupport.plan` passes the raw `event.get("when")` into `guard(Object, entity, byName)`, and `GlueIntentGenerator.guardFields` does the same for integrations/outbound — no `stringArg` stringification on either path any more.
- `CheckSupport.conditionTerms` reads a list as ANDed terms, typed per declared field.

So the ANDed list works for `generates:`, a process `trigger:`, `notifications:`, `integrations:` and `outbound:` alike; only `postings:` is status-term-only. No change owed for this finding — sorry for the noise.

